PR #4, opened 2018-01-31
schema.org for AddressWidget

code-constructor:


src/Widgets/AddressWidget/AddressWidgetComponent.js
```diff
@@ -66,3 +33,84 @@ const Logo = Scrivito.connect(() => {
...
+    );
+  }
+
+  const breakedLines = lines.reduce((array, value) => {
+    array.push(value);
+    array.push(<br key={ value } />);
+    return array;
+  }, []);
+
+  return <p>{ breakedLines }</p>;
```
apepper: How about replacing `breakedLines` with this?

```javascript
  return (
    <p>
      {
        lines.map((line, index) =>
          <React.Fragment key={ index }>
            { line }
            <br />
          </React.Fragment>
        )
      }
    </p>
  );
```

src/Widgets/AddressWidget/AddressWidgetComponent.js
```diff
@@ -66,3 +33,84 @@ const Logo = Scrivito.connect(() => {
...
+    <table>
+      <tbody>
+        { columns }
+      </tbody>
+    </table>
+  );
+}
+
+function TableColumn({ name, value }) {
+  const localization = {
```
apepper: I would extract LOCALIZATION and LINK_PREFIXES so that these objects only get created once, not for every `TableColumn` method call.

src/Widgets/AddressWidget/AddressWidgetComponent.js
```diff
@@ -66,3 +33,84 @@ const Logo = Scrivito.connect(() => {
...
+
+  return <p>{ breakedLines }</p>;
+});
+
+function Table(values) {
+  const columns = Object.keys(values).map(name => {
+    return <TableColumn key={ name } name={ name } value={ values[name] } />;
+  });
+
+  if (columns.length === 0) {
```
apepper: This does not work, since all four props are always given (but with empty values).

apepper: What do you think about the following? It also uses the newer `Object.entries` method of es7:

```javascript
function Table(props) {
  const lines = Object.entries(props).filter(([_prop, value]) => value);

  if (!lines.length) {
    return null;
  }

  return (
    <table>
      <tbody>
        {
          lines.map(([name, value]) =>
            <TableColumn key={ name } name={ name } value={ value } />)
        }
      </tbody>
    </table>
  );
}
```

src/Widgets/AddressWidget/AddressWidgetComponent.js
```diff
@@ -66,3 +33,84 @@ const Logo = Scrivito.connect(() => {
...
+  const linkPrefixes = {
+    phone: 'tel',
+    mobile: 'tel',
+    fax: 'fax',
+    eMail: 'mailto',
+  };
+
+  const prefix = linkPrefixes[name];
+
+  if (!value) {
```
apepper: If `Table` already filters out empty values (as suggested above) this return case is no longer needed.

src/Widgets/AddressWidget/AddressWidgetEditingConfig.js
```diff
@@ -25,15 +20,59 @@ Scrivito.provideEditingConfig('AddressWidget', {
...
+    },
+    eMail: {
+      title: 'E-Mail',
+    },
   },
   properties: [
     'showLogo',
-    'listItems',
     'showBorderBottom',
+    'locationName',
```
apepper: `showLogo` and `showBorderBottom` should be below the other attributes.

src/Widgets/AddressWidget/AddressWidgetClass.js
```diff
@@ -6,6 +6,18 @@ const AddressWidget = Scrivito.provideWidgetClass('AddressWidget', {
     listItems: ['widgetlist', { only: 'AddressListItemWidget' }],
```
apepper: `address` and `listItems` is no longer needed. Please remove them.

apepper: Please also adjust the `registerTextExtract` to include the newly created attributes.

src/Widgets/AddressWidget/AddressWidgetClass.js
```diff
@@ -2,16 +2,34 @@ import { registerTextExtract } from 'utils/textExtractRegistry';
...
+    locationStreetAddress: 'string',
+    locationLocality: 'string',
+    locationPostalCode: 'string',
+    locationRegion: 'string',
+    locationCountry: 'string',
+
+    phone: 'string',
+    mobile: 'string',
+    fax: 'string',
+    eMail: 'string',
```
apepper: Sorry, but I just noticed, that `E-Mail` may be correct in german, but not in english. So it's either `E-mail` or `Email`. I asked already Andreas and Antony and both do recommend `Email`. Could you go over the code-base and change all usages of `E-Mail`, `e-mail` etc. to `Email`/`email`?

Also this attribute just just be `email`.

src/Widgets/AddressWidget/AddressWidgetComponent.js
```diff
@@ -1,52 +1,33 @@
...
+import InPlaceEditingPlaceholder from 'Components/InPlaceEditingPlaceholder';
+import SchemaDotOrg from 'Components/SchemaDotOrg';
 
-const AddRow = Scrivito.connect(({ widget }) => {
-  if (!Scrivito.isInPlaceEditingActive()) { return null; }
+const LOCALIZATION = {
+  phone: 'Phone',
+  mobile: 'Mobile',
+  fax: 'Fax',
+  eMail: 'E-Mail',
```
apepper: It should be `Email`, not `E-Mail`.

src/Widgets/AddressWidget/AddressWidgetEditingConfig.js
```diff
@@ -25,15 +20,59 @@ Scrivito.provideEditingConfig('AddressWidget', {
...
+      title: 'Phone',
+    },
+    mobile: {
+      title: 'Mobile',
+    },
+    fax: {
+      title: 'Fax',
+    },
+    eMail: {
+      title: 'E-Mail',
```
apepper: Should be `Email`.

src/Components/SchemaDotOrg/dataFromAddressWidget.js
```diff
@@ -0,0 +1,22 @@
+function dataFromAddressWidget(addressWidget) {
+  return {
+    '@type': 'Place',
+    name: addressWidget.get('locationName'),
+    address: addressFromAddressWidget(addressWidget),
+    telephone: (addressWidget.get('phone') || addressWidget.get('mobile')),
+    faxNumber: addressWidget.get('fax'),
```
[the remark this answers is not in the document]
code-constructor: it's not specified as field in the schema.org definition

code-constructor:
@apepper ready for review again

src/Components/SchemaDotOrg/dataFromAddressWidget.js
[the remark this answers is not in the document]
apepper: You are right. I mixed it up with http://schema.org/PostalAddress.

PR #18, opened 2018-02-06
Explicitly require Scrivito and React.

apepper:
Leaving `window.Scrivito` still there to enable easier debugging. Also `window.React` is needed, otherwise *every* jsx component would have to manually import react.

The following ruby script has been used to do the rename:

```ruby
{
  Scrivito: "import * as Scrivito from 'scrivito';",
  React: "import * as React from 'react';",
}.each do |k, v|
  %x{ag -l -Q '#{k}.' src}.split.each do |fp|
    data = File.read(fp)

    File.open fp, 'w' do |f|
      f.write "#{v}\n"
      f.write "\n" unless (data.include? "import ")
      f.write "#{data}"
    end
  end
end
```

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,3 +1,4 @@
+import * as Scrivito from 'scrivito';
 import CurrentPageMetaData from 'Components/CurrentPageMetaData';
 import ErrorBoundary from 'Components/ErrorBoundary';
 import Footer from 'Components/Footer';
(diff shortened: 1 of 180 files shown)


apepper:
Superseeded by https://github.com/Scrivito/scrivito_example_app_js/pull/22. Thanks for your input, @dcsaszar !

PR #20, opened 2018-02-06
author image component: no image selected

marcelgo:
If the editor selects something else than an image, it currently results in an empty "space". This is replaced with a hint to the editor.


apepper:
Thanks a lot for this PR! It looks really good.

PR #22, opened 2018-02-07
No more Scrivito/React globals and relative imports

apepper:
Scrivito and React are now explicitly imported and no longer use a global `window` variable for that. Leaving `window.Scrivito` still there to enable easier debugging.

If you're updating to this PR run `npm run eslint` afterwards. It will guide you to places, where React is not imported correctly.


src/index.js
```diff
@@ -1,9 +1,9 @@
-import 'Globals';
+import 'reactPolyfills';
+import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import 'Objs/index';
-import 'Widgets/index';
+import 'Objs/';
```
dcsaszar: FYI not sure about what the spec says, but my experience is that the trailing slash is not needed.

PR #25, opened 2018-02-08
images: filter images with empty blob

marcelgo:
Solves: https://infopark.slack.com/archives/C5LGN1D42/p1518015213000769


src/Widgets/TestimonialSliderWidget/TestimonialSliderWidgetComponent.js
```diff
@@ -33,7 +34,7 @@ Scrivito.provideComponent('TestimonialSliderWidget', ({ widget }) => {
 function sliderSettings(testimonials) {
   const testimonialAuthorImages = testimonials.map(testimonial => {
     const authorImage = testimonial.get('authorImage');
-    if (!authorImage || authorImage.objClass() !== 'Image') {
+    if (isImage(authorImage)) {
```
apepper: if (!isImage(authorImage)) {

PR #33, opened 2018-02-12
FIX IE11 issue of: BlogPostPreviewList, EventOverviewWidgetComponent and JobOverviewWidgetComponent

code-constructor:
remove spreading of scrivito obj search iterators from components to support IE11


code-constructor:
@apepper ready for review

PR #35, opened 2018-02-13
Redirect obj

apepper:
redirect_obj
This PR adds a new "Redirect" obj class. This obj class allows to set a link. Once can also choose if an internal link should be opened in the Scrivito UI:

redirect properties

New "Redirect" objs can be created via the content browser:
create via content browser


apepper:
REVIEW (35bef16)

PR #38, opened 2018-02-14
Import polyfills by hand and disable babel loader auto detection

code-constructor:


apepper:
Sadly `useBuiltIns: 'usage'` did not work in our case. It appears, that the polyfills where included but executed _after_ the code that needs the polyfills. We decided to switch over to an explicit approach until we know for sure, that `useBuiltIns` is working for us.

PR #44, opened 2018-02-23
Active useBuiltIns: "usage" for babel/preset-env webpack config

code-constructor:
This PR removes the explicit imports of polyfills and let babel handle the majority of polyfill imports. We have to add three polyfills by hand which have to be available at the initial loading of the website.

1. es6.map
2. es6.set
3. es6.symbol

The first two are needed by react 16 and the last one is important to support the iterator feature of scrivito `ObjSearch` (https://www.scrivito.com/objsearch-aaa67c3157464d01).


code-constructor:
Thx for optimizing the description! 👍

PR #47, opened 2018-02-26
Add integration for intercom

code-constructor:


code-constructor:
@apepper ready for review

apepper:
FYI: Conflicts.

.eslintrc
```diff
@@ -70,6 +70,36 @@
...
+      "files": [
+        "src/Components/Intercom.js",
+      ],
+      "rules": {
+        "class-methods-use-this": 0,
+      },
+    },
+    {
+      "files": [
+        "src/intercom.js",
```
apepper: I suggest to use `.eslintignore` for this.

webpack.config.js
```diff
@@ -36,6 +36,7 @@ module.exports = (env = {}) => {
     new InlineEnvironmentVariablesPlugin('SCRIVITO_TENANT'),
     new CopyWebpackPlugin([
       { from: '../public' },
+      { from: '../src/intercom.js' },
```
apepper: Why a copy, and not something like this https://github.com/Scrivito/scrivito_example_app_js/blob/master/webpack.config.js#L74 ?

src/Objs/Homepage/HomepageEditingConfig.js
```diff
@@ -48,6 +48,10 @@ Scrivito.provideEditingConfig('Homepage', {
       description: 'Needs to be requested under your GA property settings.' +
         ' See https://analytics.google.com for details.',
     },
+    intercomAppId: {
+      title: 'Intercom App Id',
```
apepper: `Intercom App ID`.

src/intercom.js
```diff
@@ -0,0 +1 @@
+(function(){var w=window;var ic=w.Intercom;if(typeof ic==="function"){ic('reattach_activator');ic('update',intercomSettings);}else{var d=document;var i=function(){i.c(arguments)};i.q=[];i.c=function(args){i.q.push(args)};w.Intercom=i;function l(){var s=d.createElement('script');s.type='text/javascript';s.async=true;s.src='https://widget.intercom.io/widget/v4468x5h';var x=d.getElementsByTagName('script')[0];x.parentNode.insertBefore(s,x);}if(w.attachEvent){w.attachEvent('onload',l);}else{w.addEventListener('load',l,false);}}})()
```
apepper: Did you try this out on Netlify? Because it looks like, that `https://widget.intercom.io` needs to be in the CSP rules for this to work correctly.

webpack.config.js
apepper: Why a copy, and not something like this https://github.com/Scrivito/scrivito_example_app_js/blob/master/webpack.config.js#L74 ?

.eslintrc
apepper: I suggest to use .eslintignore for this.

.eslintrc
```diff
@@ -70,6 +70,36 @@
         "import/first": 0,
       },
     },
+    {
+      "files": [
+        "src/Components/Intercom.js",
+      ],
+      "rules": {
+        "class-methods-use-this": 0,
```
apepper: Is this rule needed for `Intercom.js`?

webpack.config.js
code-constructor: Because the intercom.js includes only the snippet copied from the intercom documentation in my opinion it makes no sense to let it run through babel. For google_analytics.js it is an other case, we compile it because of the `import` statement in it. 

I've tried it. The size of the intercom.js gets doubled if we include it as own entry in webpack (533B -> 981B).

.eslintrc
code-constructor: Yeah, nice idea!

.eslintrc
code-constructor: I thought the eslint error was thrown by the use of `window.`, in the intercom.js, but thats not the case. It's because ESLint suggest a static method instead of an object method for `Intercom#getIntercomAppId` (https://github.com/Scrivito/scrivito_example_app_js/pull/47/files#diff-96ce2237efc203044372d3051417e295R35). So i will change it to a static one, so we can get rid of the `.eslintrc` entry 👍

code-constructor:
@apepper ready for review again

code-constructor:
@dcsaszar ready for review again

src/Objs/Homepage/HomepageEditingConfig.js
dcsaszar: Not exactly. We're using sentence case for the labels. So it's `Intercom app ID`. 🤓

.eslintignore
```diff
@@ -0,0 +1 @@
+src/intercom.js
```
dcsaszar: * [ ] Is this an artifact? This file doesn't exist. If I'm not missing somthing, let's revert! ◀️

src/Components/Intercom.js
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+import * as Scrivito from 'scrivito';
+import Helmet from 'react-helmet';
+
+class Intercom extends React.Component {
+  static getIntercomAppId() {
```
dcsaszar: We're not using the pattern of static functions so far. AFAICT the pattern is putting helper functions after the component (see e.g. SchemaDotOrg).

* [ ] Please don't use static functions for better consistency.

src/Components/Intercom.js
```diff
@@ -0,0 +1,54 @@
...
+    const rootPage = Scrivito.Obj.root();
+
+    if (!rootPage) {
+      return;
+    }
+
+    return rootPage.get('intercomAppId');
+  }
+
+  static bindFakeIntercom() {
```
dcsaszar: As discussed, we shouldn't name it "fake". The "snippet API" doesn't indicate this.
* [ ] find a better name.

src/Components/Intercom.js
```diff
@@ -0,0 +1,64 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isIntercomInstalled: false,
+    };
+  }
+
+  componentDidMount() {
+    Scrivito.load(() => getIntercomAppId()).then(intercomAppId => {
```
dcsaszar: * [ ] FYI: the anonymous function indirection is not necessary. `Scrivito.load(getIntercomAppId).then`

PR #53, opened 2018-03-16
show widget section for medium sized headers

code-constructor:
Allow medium sized headers to show the widget list.

bildschirmfoto 2018-03-16 um 14 57 50

@agessler can you pls have a look if the design is good or what we can do to change the padding?


agessler:
@dcsaszar Please review again.

PR #54, opened 2018-03-23
Less misleading description for publish date

dcsaszar:
The old description can be confused with a "valid from" date.


apepper:
Thank you for your contribution!

PR #57, opened 2018-04-03
React 16.3

apepper:
This PR updates the example app to the latest and greatest react version 16.3. It also uses some of its new features. For details see https://reactjs.org/blog/2018/03/29/react-v-16-3.html.


src/Components/Navigation/Search.js
```diff
@@ -4,9 +4,12 @@ import * as Scrivito from 'scrivito';
 const SearchBox = Scrivito.connect(class extends React.Component {
   constructor(props) {
     super(props);
+
     this.state = {
       q: '',
     };
+
+    this.inputRef = React.createRef();
```
dcsaszar: *  [ ] The React example is `this.textInput = React.createRef();`. I don't think suffixing the names with `Ref` is helpful. We should keep the names as they are.

src/Objs/SearchResults/SearchInput.js
```diff
@@ -10,10 +10,13 @@ class SearchInput extends React.Component {
     };
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (this.props.params.q !== nextProps.params.q) {
-      this.setState({ q: nextProps.params.q });
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (!prevState.params || !nextProps.params || prevState.params.q === nextProps.params.q) {
```
dcsaszar: * [ ] I don't think the change in behavior is intentional: It is no longer possible to clear `state.q` by passing no `params`.

src/Objs/SearchResults/SearchInput.js
```diff
@@ -10,10 +10,13 @@ class SearchInput extends React.Component {
...
+      // Return null to indicate no change to state.
```
dcsaszar: * [ ] This is documented behavior, we don't need that comment.

src/Components/Navigation/Search.js
apepper: Quote from https://reactjs.org/blog/2018/03/29/react-v-16-3.html#createref-api :

```jsx
class MyComponent extends React.Component {
  constructor(props) {
    super(props);

    this.inputRef = React.createRef();
  }

  render() {
    return <input type="text" ref={this.inputRef} />;
  }

  componentDidMount() {
    this.inputRef.current.focus();
  }
}
```

I actually think that the `Ref` suffix makes it clearer, especially for users of that variable.

apepper:
REVIEW (f46d189)

src/Components/Navigation/Search.js
dcsaszar: I see, the docs aren't consistent. Fine with me. 👍

src/Objs/SearchResults/SearchInput.js
```diff
@@ -11,12 +11,10 @@ class SearchInput extends React.Component {
   }
 
   static getDerivedStateFromProps(nextProps, prevState) {
-    if (!prevState.params || !nextProps.params || prevState.params.q === nextProps.params.q) {
-      // Return null to indicate no change to state.
-      return null;
-    }
+    const newQ = (nextProps.params || {}).q;
```
dcsaszar: FYI could also be written as
`const newQ = nextProps.params && nextProps.params.q;`
or even
`const q = nextProps.params && nextProps.params.q;`

PR #58, opened 2018-04-05
Sitemap.xml

apepper:
This PR generates a static `sitemap.xml`. See it live under https://sitemap-xml--scrivito-example.netlify.com/sitemap.xml .

Due to url redirects this is currently only working for full netlify deployments (e.g. via github deploy hooks) and _not_ using zip deployments.


src/utils/sitemapXml.js
```diff
@@ -0,0 +1,41 @@
...
+import formatDate from './formatDate';
+
+const OBJ_CLASSES_TO_INDEX = [
+  'Author',
+  'Blog',
+  'BlogPost',
+  'Event',
+  'Homepage',
+  'Job',
+  'LandingPage',
```
dcsaszar: In some websites a "Landing page" is a page outside of the page hierarchy, with no connection to the main website. In these scenarios you don't want landings to be part of the sitemap.

* [ ] Please check if this matches the example app usage of landing pages. If so, we should exclude `LandingPage` from the list.

add-sitemap-to-redirects-webpack-plugin.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+}
+
+function emit(compilation, callback) {
+  const tenant = process.env.SCRIVITO_TENANT;
+  const url = process.env.DEPLOY_PRIME_URL;
+
+  if (url && tenant) {
+    const redirects = compilation.assets['_redirects'].source().toString();
+    const patchedRedirects = redirects.replace(/^\/\* \/.+$/m, rewriteRule(tenant, url));
```
dcsaszar: This took me very long to figure out. The idea seems to be "insert before the 1st wildcard".
* [ ] Can we make this easier to read?
My proposal is having a placeholder/commented line at the right spot, so we can do a more simple replace.

add-sitemap-to-redirects-webpack-plugin.js
```diff
@@ -0,0 +1,32 @@
...
+    compilation.assets['_redirects'] = {
+      source: () => patchedRedirects,
+      size: () => patchedRedirects.length,
+    };
+  }
+
+  callback();
+}
+
+function rewriteRule(tenant, url) {
```
dcsaszar: Some issues here:

* It's actually *two* rules
* `/_sitemaphtml /_sitemap.html 200` can be hardcoded
* The readability can be improved by having blank lines, like:
```js
return `
/rule1
/rule2
`;
```

* [ ] Try improving readability

src/utils/sitemapXml.js
apepper: Isn't a `LandingPage` built exactly for the "specialized search via google" use-case? So if should be listed here. If the landing pages is for something else than the google index, then "Should this page be indexed?" can be set to "No" for that landing page.

apepper:
REVIEW (df5ea03)

add-sitemap-to-redirects-webpack-plugin.js
```diff
@@ -0,0 +1,37 @@
...
+  };
+}
+
+function emit(compilation, callback) {
+  const tenant = process.env.SCRIVITO_TENANT;
+  const url = process.env.DEPLOY_PRIME_URL;
+
+  if (url && tenant) {
+    const previousRedirects = compilation.assets['_redirects'].source().toString();
+    const redirects = addSitemapTRedirects(tenant, url, previousRedirects);
```
dcsaszar: T => To 😁

PR #63, opened 2018-04-24
Package dev dependencies are now dependencies

dcsaszar:
* According to https://github.com/webpack/webpack/issues/520#issuecomment-331689542, all deps should reside in `dependencies`
* This also improves compatibility with the default behavior of `npm install x`

There is some debate about where dependencies go. IMO a good rationale is the summary of https://github.com/webpack/webpack/issues/520#issuecomment-331689542:

> Webpack projects should be considered build-type projects first, and not production-type projects, unless webpack is used in production. Packages needed during production goes in `dependencies`, and packages used just to develop the project goes in `devDependencies`. If our production environment needs to build (run webpack), then those dependencies must reside in `dependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "repository": {
     "private": true
   },
-  "devDependencies": {
+  "dependencies": {
     "@babel/core": "^7.0.0-beta.46",
     "@babel/preset-env": "^7.0.0-beta.46",
     "@babel/preset-react": "^7.0.0-beta.46",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,13 +4,13 @@
   "license": "MIT",
   "engines": {
     "node": ">=8.0.0",
-    "npm": ">=5.0.0"
```
apepper: Please to do not yet force `npm` 6. The latest released `node` version only contains `npm` 5.

dcsaszar:
Force pushed. REVIEW

PR #73, opened 2018-07-06
Scrivito 1.2.0

apepper:
Scrivito `1.2.0` is available. See [Scrivito `1.2.0` blog post](https://www.scrivito.com/scrivito-js-sdk-1-2-0-is-available-5419bde4289207f4) for details.

This PR updates to the latest scrivito version and also updates most npm packages. It also reenables editing of different attributes in the page/widget properties:

* Name, description and image of an `Author`.
* Title of an `Event` and a `Job`.
* Text of a `TextWidget`.

Also the testimonial and author of a slider of the `TestimonialSliderWidget` can now also be edited inplace.


src/Widgets/TestimonialSliderWidget/TestimonialSliderWidgetComponent.js
```diff
@@ -16,12 +16,18 @@ Scrivito.provideComponent('TestimonialSliderWidget', ({ widget }) => {
           testimonials.map(testimonial =>
             <div key={ testimonial.id() }>
               <h1 className="quote-headline text-center">&quot;</h1>
-              <p className="h4 text-center">
-                { testimonial.get('testimonial') }
-              </p>
-              <p className="small text-center">
-                { testimonial.get('author') }
-              </p>
+              <Scrivito.ContentTag
```
dcsaszar: * [ ] FYI not the focus of this PR, but to make this feature complete, the image should also be editable in-place

apepper: That will be hard to do, since the Slick Slider component used already has a click event on the images - to move to the slider. I don't think it is worth it to build code, that "fights" with Slick-Slider who can use the click or not.

apepper:
REVIEW (d03041d)

* Updated title of a job.

PR #76, opened 2018-07-13
Use prettier defaults

dcsaszar:
The goal of #75 was
> to no longer have to discuss styling again.

As long as we use custom options, we still have to argue. Especially the column width is a hot topic I don't like to see discussed here.
If we use the defaults, all discussions can go to https://github.com/prettier/prettier/issues.

This PR removes the custom prettier options, and reformats the code using the defaults . `npx prettier --write src/**/*.js`

The code looks pretty pretty to me. ✨


package.json
```diff
@@ -25,6 +25,7 @@
     "ent": "^2.2.0",
     "eslint": "^4.19.1",
     "eslint-config-airbnb": "^16.1.0",
+    "eslint-config-prettier": "^2.9.0",
```
apepper: We already have `eslint-plugin-prettier`. Why do we need to also have `eslint-config-prettier`?

PR #83, opened 2018-08-07
Changed text-align of links in nav

karminkarmen:
https://trello.com/c/QkmEwbJn/340-navigation-on-mobile-for-pages-without-path-look-different

`navbar-nav` object inherits text-align from `medium-height` element, added specific `text-align` value for links in the nav


agessler:
ok

PR #86, opened 2018-08-08
Added margin to tags under a blog post

karminkarmen:
https://trello.com/c/oJ0X0umU/348-tag-list-under-a-blog-post-looks-broken-if-there-are-more-than-one-tag

Added new class `margin-right` and used it for tags under blog post


apepper:
See it in action on https://deploy-preview-86--scrivito-example.netlify.com/cupidatat-ad-ad-velit-id-80bee898db66b040

PR #87, opened 2018-08-08
Striptags 3

apepper:
Use the latest version of `striptags`. It needs to be compiled by babel, but should *not* be enriched with polyfills, since that will break webpack with the following message:

```
"export 'default' (imported as 'striptags') was not found in 'striptags'
 @ ./utils/textExtract.js
 @ ./Objs/SearchResults/SearchResultItem.js
 @ ./Objs sync \.js$
 @ ./Objs/index.js
 @ ./sitemap.js
```

Thanks to the [`overrides`](https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages#selective-compilation-with-overrides) option it is possible to disable the given `@babel/preset-env` rule.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9471,9 +9471,9 @@
       "integrity": "sha1-PFMZQukIwml8DsNEhYwobHygpgo="
     },
     "striptags": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/striptags/-/striptags-2.2.1.tgz",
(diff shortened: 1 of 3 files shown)


apepper:
FYI: I tested the preview deployment on IE11 using Sauce-Labs. Search results look good!

apepper:
FYI: Package sizes stays the same.

PR #90, opened 2018-08-08
Added chrome 41 to targets in webpack config

karminkarmen:
https://trello.com/c/TVAd8pU7/335-add-chrome-41-as-babel-env-target

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -93,7 +93,9 @@ module.exports = (env = {}) => {
                       modules: false,
                       shippedProposals: true,
                       useBuiltIns: "usage",
-                      targets: { browsers: ["last 2 versions"] },
+                      targets: {
+                        browsers: ["Chrome >= 41", "last 2 versions"],
+                      },
                     },
                   ],
                 ],


webpack.config.js
```diff
@@ -93,7 +93,7 @@ module.exports = (env = {}) => {
                       modules: false,
                       shippedProposals: true,
                       useBuiltIns: "usage",
-                      targets: { browsers: ["last 2 versions"] },
+                      targets: { browsers: ["last 2 versions"], chrome: 41 },
```
apepper: I think the format is more like this:

```js
targets: {
  browsers: [
    "Chrome >= 41",
    "last 2 versions",
  ],
}
```

Can you run `rm -rf tmp/ && npm run build` with `debug: true` for `@babel/preset-env` to see the effect? `@babel/preset-env` should log out, which chrome version it now targets.

karminkarmen:
REVIEW (f277c92) /cc @apepper

webpack.config.js
karminkarmen: It log out version 41 of chrome in targets. I checked also which version of chrome the google search spiders currently uses and it's 41.

PR #96, opened 2018-08-20
Added new file with whitelist

karminkarmen:
https://trello.com/c/4ccNxs08/365-split-of-whitelisted-domains-for-csp-into-its-one-file-that-gets-processed-on-deploy

Added file with listed whitelist links.


apepper:
Talked about it in person. Please resubmit, when ready.

PR #100, opened 2018-08-27
Cookie consent policy bottom banner

karminkarmen:
https://trello.com/c/JJE0AFZh/369-cookie-consent-policy-popup

Created cookie consent policy bar using react-cookie-consent npm package.


src/Components/CookieConsent.js
```diff
@@ -0,0 +1,50 @@
...
+  const cookiesLinkTitle = cookieLink && cookieLink.title();
+
+  if (!cookieLink) {
+    return null;
+  }
+
+  return (
+    <CookieConsent
+      buttonText="accept"
+      disableStyles="true"
```
apepper: It should be `disableStyles={true}`. Otherwise you'll get a js console warning:

```
Warning: Failed prop type: Invalid prop `disableStyles` of type `string` supplied to `CookieConsent`, expected `boolean`
```

src/Components/CookieConsent.js
```diff
@@ -0,0 +1,50 @@
...
+      containerClasses="cookie-box flex-row d-lg-flex d-sm-flex"
+      contentClasses="cookie-content flex-row d-flex"
+      buttonClasses="btn btn-primary cookie-button ml-auto"
+    >
+      <div className="cookie-img-box">
+        <img
+          className="cookie-img"
+          src="../../src/assets/images/iconmonstr-candy-6.svg"
```
apepper: Please do not hard-code paths to files. Instead please use imports as provided by the webpack [`file-loader`](https://github.com/webpack-contrib/file-loader) plugin. It works like this:

You simply import the file like this:

```js
import cookieConsentIcon from "../assets/images/iconmonstr-candy-6.svg";
```

And than you can use the filename as this:

```jsx
<img className="cookie-img" src={cookieConsentIcon} />
```

Please note, that at least on my machine I had to restart the webpack server after the change.

apepper: Please also change the file name of `iconmonstr-candy-6.svg`. All other svg files are named snake case. I suggest a name like `cookie_consent_icon.svg`.

Also one more thing: What is the source of the icon? Is it public domain? Or do we need to give credit for it?

src/Components/CookieConsent.js
```diff
@@ -0,0 +1,50 @@
...
+ * @return {null}
+ */
+function CookieConsentBox() {
+  const root = Scrivito.Obj.root();
+  const cookieLink = root.get("cookiesPrivacyPolicyLink");
+  const cookiesLinkTitle = cookieLink && cookieLink.title();
+
+  if (!cookieLink) {
+    return null;
+  }
```
apepper: You can also optimize the flow of code:

```js
  const cookieLink = root.get("cookiesPrivacyPolicyLink");
  if (!cookieLink) {
    return null;
  }

  const cookiesLinkTitle = cookieLink.title() || "Learn more »";
```

Later on, you would only use `cookiesLinkTitle` in the jsx part.

src/Components/CookieConsent.js
```diff
@@ -0,0 +1,50 @@
...
+        />
+      </div>
+      <div>
+        <div className="cookie-box-title">Cookies policy</div>
+        <div className="cookie-box-content">
+          Our website uses cookies to analyse how the site is used and to ensure
+          your experience is consistent between visits.
+          <a
+            className="cookie-box-link"
+            href={cookieLink}
```
apepper: This does not work and results in a `[object Object]` href, since you are including a `Scrivito.Link` object here. Please make use of the [`Scrivito.urlFor`](https://www.scrivito.com/js-sdk/urlFor) for a valid href value.

src/Components/CookieConsent.js
```diff
@@ -0,0 +1,50 @@
+import * as React from "react";
+import * as Scrivito from "scrivito";
+import CookieConsent from "react-cookie-consent";
+
+/**
+ * @return {null}
+ */
```
apepper: Please remove the comments, since they are not accurate and don't provide much value.

src/Components/CookieConsent.js
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
+import * as Scrivito from "scrivito";
+import ReactCookieConsent from "react-cookie-consent";
+import cookieConsentIcon from "../assets/images/cookie_consent_icon.svg";
+
+function CookieConsent() {
+  const root = Scrivito.Obj.root();
+  const cookieLink = root.get("cookieConsentLink");
```
apepper: If `root` is `null` this will fail.

For details see https://www.scrivito.com/obj-root-f3e15c4a58f46c68

apepper: Also one should usually try to name one thing with as few names as possible. Therefore I suggest this:

```js
const cookieConsentLink = root.get("cookieConsentLink");
```

src/Objs/Homepage/HomepageEditingConfig.js
```diff
@@ -62,6 +62,11 @@ Scrivito.provideEditingConfig("Homepage", {
         "Removes the header navigation and only centers the logo instead. Default: No",
       values: [{ value: "yes", title: "Yes" }, { value: "no", title: "No" }],
     },
+    cookieConsentLink: {
+      title: "Cookie privacy policy link",
```
apepper: I suggest to change the title to `Cookie consent link`.

See https://github.com/Scrivito/scrivito_example_app_js/pull/100#discussion_r213265683 for details.

src/assets/images/iconmonstr-candy-6.svg
```diff
@@ -0,0 +1 @@
+<svg width="50" height="50" xmlns="http://www.w3.org/2000/svg"  fill="white" fill-rule="evenodd" clip-rule="evenodd"><path transform="scale(2 2)" d="M23.999 12.149c-.049 3.834-1.893 7.223-4.706 9.378-1.993 1.53-4.485 2.449-7.198 2.473-6.464.057-12.051-5.107-12.095-12 3.966 1.066 7.682-1.993 6-6 4.668.655 6.859-2.389 6.077-6 6.724.064 11.999 5.542 11.922 12.149zm-15.576-4.123c-.065 3.393-2.801 5.868-6.182 6.166 1.008 4.489 5.015 7.807 9.759 7.807 5.262 0 9.576-4.072 9.97-9.229.369-4.818-2.755-9.357-7.796-10.534-.277 2.908-2.381 5.357-5.751 5.79zm5.077 8.974c.828 0 1.5.672 1.5 1.5s-.672 1.5-1.5 1.5-1.5-.672-1.5-1.5.672-1.5 1.5-1.5zm-5.5-2.853c1.104 0 2 .896 2 2s-.896 2-2 2-2-.896-2-2 .896-2 2-2zm10-2.147c1.104 0 2 .896 2 2s-.896 2-2 2-2-.896-2-2 .896-2 2-2zm-5 0c.552 0 1 .448 1 1s-.448 1-1 1-1-.448-1-1 .448-1 1-1zm2.5-5c.828 0 1.5.672 1.5 1.5s-.672 1.5-1.5 1.5-1.5-.672-1.5-1.5.672-1.5 1.5-1.5zm-12.5 0c.552 0 1 .448 1 1s-.448 1-1 1-1-.448-1-1 .448-1 1-1zm-1.5-4c.828 0 1.5.672 1.5 1.5s-.672 1.5-1.5 1.5-1.5-.672-1.5-1.5.672-1.5 1.5-1.5zm6-2c.828 0 1.5.672 1.5 1.5s-.672 1.5-1.5 1.5-1.5-.672-1.5-1.5.672-1.5 1.5-1.5zm-3.5-1c.552 0 1 .448 1 1s-.448 1-1 1-1-.448-1-1 .448-1 1-1z"/></svg>
```
apepper: This file is now obsolete and can be removed.

karminkarmen:
REVIEW (235e679) /cc @apepper  Added extra styles for editing mode and for media queries, changed icon (license need a link see -> https://support.flaticon.com/hc/en-us/articles/207248209-How-I-must-insert-the-attribution-)

karminkarmen:
REVIEW (235e679) /cc @apepper  
* Added extra styles for editing mode and for media queries, 
* Changed icon (license need a link -> https://support.flaticon.com/hc/en-us/articles/207248209-How-I-must-insert-the-attribution-)

karminkarmen:
REVIEW (9b68dd4) /cc @apepper

src/Components/CookieConsent.js
```diff
@@ -0,0 +1,47 @@
...
+import cookieConsentIcon from "../assets/images/cookie_consent_icon.svg";
+
+function CookieConsent() {
+  const root = Scrivito.Obj.root();
+
+  if (!root) {
+    return null;
+  }
+
+  const cookieConsentLink = root.get("cookieConsentLink");
```
apepper: This is missing:

```js
if (!cookieConsentLink) {
  return null;
}
```

Otherwise the component will fail, if no `cookieConsentLink` attribute is set.

karminkarmen:
REVIEW (3c7aebc) /cc @apepper

karminkarmen:
REVIEW (3c7aebc) /cc @agessler

karminkarmen:
REVIEW (f04f126) /cc @apepper Added linguistic corrections

agessler:
@apepper Looks good...

PR #104, opened 2018-08-30
Extended headers

karminkarmen:
https://trello.com/c/h3eitIb1/374-additional-headers-so-that-securityheaderscom-is-happy

Added new headers:
`Feature-Policy`
`Referrer-Policy`
`X-Content-Type-Options`
`X-XSS-Protection`
`X-Frame-Options`

Result: https://securityheaders.com/?q=https%3A%2F%2Fsparkling-prune-scrivito-0728.scrivito.net%2F&followRedirects=on

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -7,3 +7,8 @@
 
 /*
   Content-Security-Policy: CSP-DIRECTIVES-PLACEHOLDER;
+  X-frame-options: sameorigin
+  X-XSS-Protection: 1; mode=block
+  X-Content-Type-Options: nosniff
+  Referrer-Policy: strict-origin-when-cross-origin;
+  Feature-Policy: payment 'self'; geolocation 'self'
--- a/public/_headers
+++ b/public/_headers
@@ -11,4 +11,3 @@
   X-XSS-Protection: 1; mode=block
   X-Content-Type-Options: nosniff
   Referrer-Policy: strict-origin-when-cross-origin
-  Feature-Policy:
--- a/public/_headers
+++ b/public/_headers
@@ -10,5 +10,5 @@
   X-frame-options: sameorigin
   X-XSS-Protection: 1; mode=block
   X-Content-Type-Options: nosniff
-  Referrer-Policy: strict-origin-when-cross-origin;
-  Feature-Policy: payment 'self'; geolocation 'self'
+  Referrer-Policy: strict-origin-when-cross-origin
+  Feature-Policy:


public/_headers
```diff
@@ -7,3 +7,8 @@
 
 /*
   Content-Security-Policy: CSP-DIRECTIVES-PLACEHOLDER;
+  X-frame-options: sameorigin
+  X-XSS-Protection: 1; mode=block
+  X-Content-Type-Options: nosniff
+  Referrer-Policy: strict-origin-when-cross-origin;
```
apepper: When you're navigating to the scrivito UI the following warning is shown in the console:

```
Failed to set referrer policy: The value 'strict-origin-when-cross-origin;' is not one of 'no-referrer', 'no-referrer-when-downgrade', 'origin', 'origin-when-cross-origin', 'same-origin', 'strict-origin', 'strict-origin-when-cross-origin', or 'unsafe-url'. The referrer policy has been left unchanged.
```
It looks like simply removing the trailing `;` should fix the issue.

public/_headers
```diff
@@ -7,3 +7,8 @@
...
+  Feature-Policy: payment 'self'; geolocation 'self'
```
apepper: The example app is the foundation for new scrivito apps. I can't foresee which features the developer wants to use or how she/he will use them. E.g. if someone will incorporate a payment iframe or a payment widget.

Therefore I suggest to just remove the `Feature-Policy` since I don't see a good default value, that will not hinder our customers, if they try to implement such a feature.

Or we could simply set it empty:

```
Feature-Policy:
```

This way we would still pass the security headers test:

> To further the exposure of the header I will be adding detection to Security Headers. The header will not have a weight to impact the grade but it will be flagged as required. I can start to track utilisation and make people aware of its existence to drive adoption!

Source: https://scotthelme.co.uk/a-new-security-header-feature-policy/

karminkarmen:
REVIEW (3c7aebc) /cc @apepper

public/_headers
```diff
@@ -7,3 +7,8 @@
 
 /*
   Content-Security-Policy: CSP-DIRECTIVES-PLACEHOLDER;
+  X-frame-options: sameorigin
+  X-XSS-Protection: 1; mode=block
+  X-Content-Type-Options: nosniff
+  Referrer-Policy: strict-origin-when-cross-origin
+  Feature-Policy:
```
apepper: Netlify seems to prune empty headers:

missing header

Then let us just remove the "Feature-Policy" header for now.

As seen on https://securityheaders.com/?q=https%3A%2F%2Fdeploy-preview-104--scrivito-example.netlify.com&hide=on&followRedirects=on

karminkarmen:
REVIEW (f9b7408) /cc @apepper

PR #108, opened 2018-09-04
Changed take(1) on first()

karminkarmen:
https://trello.com/c/G50MIOcA/384-replace-take1-with-first

Changed method `take(1)` on `first()` according to new scrivito version (1.3.0).
See documentation for `first()` method -> https://www.scrivito.com/js-sdk/ObjSearch-first

### DIFF
--- a/src/Components/BlogPost/BlogPostNavigation.js
+++ b/src/Components/BlogPost/BlogPostNavigation.js
@@ -34,7 +34,7 @@ const BlogPostNextLink = Scrivito.connect(({ currentBlogPost }) => {
   const currentDate = currentBlogPost.get("publishedAt");
 
   // find greater than publishedAt
-  const [newerPost] = Scrivito.getClass("BlogPost")
+  const newerPost = Scrivito.getClass("BlogPost")
     .where("publishedAt", "isGreaterThan", currentDate)
     .order("publishedAt", "asc")
     .first();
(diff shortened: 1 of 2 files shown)


src/Components/BlogPost/BlogPostNavigation.js
```diff
@@ -37,7 +37,7 @@ const BlogPostNextLink = Scrivito.connect(({ currentBlogPost }) => {
   const [newerPost] = Scrivito.getClass("BlogPost")
```
apepper: It now needs to be `const newerPost = Scrivito.getClass("BlogPost")`, since `first()` will already return only one item.

src/Components/BlogPost/BlogPostNavigation.js
```diff
@@ -59,7 +59,7 @@ const BlogPostPreviousLink = Scrivito.connect(({ currentBlogPost }) => {
     .andNot("id", "equals", currentBlogPost.id())
     .andNot("publishedAt", "isGreaterThan", currentDate)
     .order("publishedAt", "desc")
-    .take(1);
```
apepper: It now needs to be `const olderPost = Scrivito.getClass("BlogPost")`, since `first()` will already return only one item.

src/Components/Navigation/Search.js
```diff
@@ -70,7 +70,7 @@ const SearchBox = Scrivito.connect(
 function oldestSearchResultsPage() {
   return Scrivito.Obj.where("_objClass", "equals", "SearchResults")
     .order("_createdAt")
-    .take(1)[0];
+    .first()[0];
```
apepper: It now should be `.first();`

PR #110, opened 2018-09-04
Added 'before footer' notice

karminkarmen:
https://trello.com/c/8mKBIMrG/376-footer-notice-that-footer-starts-here

Added info before the footer, that it'll look the same on every page.
The notice was consulted with Andreas.


apepper:
This PR adds a note in edit mode right before the footer:

footer note

PR #111, opened 2018-09-05
Bugfix: Autoreloading works again

apepper:
Since the introduction of `ExtendCspHeadersWebpackPlugin` `npm start` blocks on auto recompile.

This PR does two thinks:

* Switch from `tapAsync` to `tapPromise` because it helps better to surface errors.
* Skip `ExtendCspHeadersWebpackPlugin` if `_headers` or `_headersCsp.json` is missing. This is for example the case during a auto recompile, when running `webpack-dev-server`.

### DIFF
--- a/ExtendCspHeadersWebpackPlugin.js
+++ b/ExtendCspHeadersWebpackPlugin.js
@@ -12,15 +12,17 @@ const builder = require("content-security-policy-builder");
 function ExtendCspHeadersWebpackPlugin() {
   return {
     apply: compiler => {
-      compiler.hooks.emit.tapAsync("ExtendCspHeadersWebpackPlugin", emit);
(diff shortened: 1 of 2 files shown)


ExtendCspHeadersWebpackPlugin.js
```diff
@@ -19,9 +19,13 @@ function ExtendCspHeadersWebpackPlugin() {
 
 function emit(compilation) {
   return new Promise(resolve => {
-    const csp = generateCsp(compilation.assets);
-    extendHeaders(compilation.assets, csp);
-    resolve();
+    const assets = compilation.assets;
+    if (!assets["_headers"] || !assets["_headersCsp.json"]) {
```
dcsaszar: Now that we are checking the individual assets in the caller anyways,
we should only pass the needed assets to the called functions.
This prevents duplication of the `assets[xyz]` code by making the called functions more dumb.

PR #126, opened 2018-10-04
Redirect obj bugfix

apepper:
This PR reverts https://github.com/Scrivito/scrivito_example_app_js/pull/105. `Scrivito.navigateTo` is not good enough in this case, because it does do a browser location push instead of a replace. This means, that a "navigate back click" does bring you back to the redirect obj, which then will redirect you again - quite annoying and also noticed by our customers.

This PR also includes two improvements for Redirect Objs:

* ~~When in edit mode the redirect does not take place. This allows the editor to change page properties of a redirect without having to go to the Content Browser.~~ **EDIT**: Did remove the functionality for now.
* A redirect obj now also has a title. This is useful, if someone includes the redirect obj in the navigation. Then it should be possible to edit the title.


apepper:
REVIEW (2237f13)

PR #129, opened 2018-10-11
Replaced extract-text-webpack-plugin with mini-css-extract-plugin

karminkarmen:
This PR changed plugin from `extract-text-webpack-plugin` to `mini-css-extract-plugin` in webpack configuration.


webpack.config.js
```diff
@@ -116,19 +116,16 @@ module.exports = (env = {}) => {
...
-                  minimize: isProduction,
-                },
-              },
-              {
-                loader: "sass-loader",
+          use: [
+            {
+              loader: MiniCssExtract.loader,
+              options: {
+                minimize: isProduction,
```
apepper: AFAIK `mini-css-extract-plugin` does not have a `minimize` option. On their github page they mention the `optimize-css-assets-webpack-plugin` package for that (see [1] for details).

Can you try out `optimize-css-assets-webpack-plugin` and measure if it does have any effect on the resulting `index.css` file size? If it does not have any effect than please just remove the `minimize` option.

[1] https://github.com/webpack-contrib/mini-css-extract-plugin#minimizing-for-production

karminkarmen: I added `optimize-css-assets-webpack-plugin` but there's no effect on the `index.css` file size. You can see changes 58e7f3c

karminkarmen: I tried add `optimize-css-assets-webpack-plugin` with few configurations but with no effect on `index.css`. Only `uglifyjs` affects the size of `index.js` (result 894 kB from 898 kB)

karminkarmen:
REVIEW (922992b) /cc @apepper `index.css` 222 kB -> 218 kB

PR #143, opened 2018-11-21
Prerender security

apepper:
Given the following `rerenderContent`:

```js
function prerenderContent() {
  return new Promise(resolve => {
    resolve([
      { fileName: "../bad.html", fileContent: "should not be there" },
      { fileName: "/good.html", fileContent: "isFine" },
      { fileName: "/also/good.html", fileContent: "isFineToo" },
      { fileName: "google_analytics.js", fileContent: "OVERWRITE!!" },
    ]);
  });
}
```

The file `bad.html` should not be stored, since it could "escape" the `TARGET_DIR`. Also `google_analytics.js` should not be overwritten, since it already exists in `build/`.

If one of the two cases is detected, it is not stored and a warning is logged. I'm not sure, if we should fail the build, if this happens.

### DIFF
--- a/storePrerenderedContent.js
+++ b/storePrerenderedContent.js
@@ -1,5 +1,6 @@
 const express = require("express");
 const fse = require("fs-extra");
+const path = require("path");
 const puppeteer = require("puppeteer");
 
 const TARGET_DIR = "buildPrerendered";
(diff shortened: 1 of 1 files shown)


apepper:
REVIEW (224a00a)

PR #149, opened 2018-12-07
[prerender] storePrerenderedContent now extends _redirects with upper…

apepper:
…case urls.

This is a workaround to netlifys url normalization. If one has a permalink with an upcase such as "myPermalink", netlify would previously redirect to "/mypermalink". Once Scrivito JS is started it would no longer recognize that permalink, since permalinks are case-sensitive. With an explicit redirect entry in "_redirects", Netlify does not "normalize" the given URL to lower-case.

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -8,4 +8,5 @@
 
 # Single Page Application routes
 /404 /catch_all_index.html 404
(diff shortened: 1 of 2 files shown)


apepper:
FYI: Travis is green - https://travis-ci.org/Scrivito/scrivito_example_app_js/builds/464940087.

PR #154, opened 2018-12-10
[prerender] Log url of failed obj.

apepper:
Wished by Andreas.

FYI: `Scrivito.urlFor` is called outside of a `load` context. But at the time of the call `Scrivito.urlFor` should have already all needed data to not complain about incomplete data.


dcsaszar:
I was too fast on the draw: One suggestion: because I know Andreas had the issue with the `index.html` (which was missing), it would be great to see a hint about `index.html` in the log.

PR #156, opened 2018-12-13
Extract comments from bundle

dcsaszar:
Saves 2.15 k

See https://github.com/webpack-contrib/terser-webpack-plugin#extractcomments

IMO the licence is still distributed with the software, but if you are unsure about legal implications, feel free to close.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -109,6 +109,7 @@ function webpackConfig(env = {}) {
       minimizer: [
         new TerserPlugin({
           cache: true,
+          extractComments: true,
           parallel: true,
           terserOptions: { ecma: 5 },
         }),


apepper:
How come, that `index.js.LICENSE` is `273 KiB` (see https://travis-ci.org/Scrivito/scrivito_example_app_js/jobs/467474078#L549)? Also there is a `index.css.js.LICENSE`, but not a `index.css.LICENSE` file.

dcsaszar:
Force pushed. I misread the difference between `extractComments: true` and ``extractComments: "all"`. `true` is the right choice here.

apepper:
FYI: I rebase the commit to the latest `master`.

PR #157, opened 2018-12-17
Integrate reference restriction feature

phorsuedzie:


phorsuedzie:
I noticed some inconsistencies in use of `'`vs. `"`. Make it consistent?

apepper:
We need Scrivito 1.5.0 first, in order for it to work. Please resubmit, once we do have `scrivito` `1.5.0` in the codebase.

PR #167, opened 2019-01-29
Limit reference/referencelist attributes to obj classes.

apepper:
Make use of the new `scrivito` `1.5.0` feature of limiting selection in the content browser:

### Default filter (unmodified)

contentbrowser_default

How to reach: Open the content browser via the sidebar.

### Image filter (only one valid objClass)
contentbrowser_single_restriction

How to reach: In edit mode click on an ImageWidget.

### Background filter (multiple valid objClasses)

contentbrowser_multiple_restrictions

How to reach: In edit mode click on the page properties of a Page. Then click "Background".

Superseeds #164.


src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,129 @@
...
-            Homepage: {
-              title: "Homepage",
-              icon: "inbox",
-            },
-            Blog: {
-              title: "Blog",
-              icon: "pen",
+  filters: filterContext => {
+    if (filterContext._validObjClasses && filterContext._validObjClasses[0]) {
+      const validObjClasses = filterContext._validObjClasses;
```
phorsuedzie: This const may be moved ahead of the condition.

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,129 @@
...
             },
-            SearchResults: {
-              title: "Search results",
-              icon: "lens",
+          },
+          _modification: modificationFilter,
+        };
+      }
+
+      return {
```
phorsuedzie: I suggest to move this simple, special(?) case ahead of the generic one.

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,129 @@
...
-        Video: {
-          title: "Videos",
-          icon: "video",
-          field: "_objClass",
-          value: "Video",
+        _modification: modificationFilter,
+      };
+    }
+
+    const pages = [
```
phorsuedzie: Extract all code into a (clearly non-parametrized) `return defaultFiltersDefinition()` for better code climate?

AFAICS it is even a constant object - not even `new Date()` involved - similar to `_modificationFilter`.

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,129 @@
...
+          ...{ Video: descriptionForObjClass("Video") },
         },
       },
+      _modification: modificationFilter,
+    };
+  },
+});
+
+function descriptionForObjClass(objClass) {
+  const descriptions = {
```
phorsuedzie: Like `pages`, this const could be defined outside of the function.

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,129 @@
...
+            options: {
+              All: {
+                title: "All",
+                icon: "folder",
+                field: "_objClass",
+                value: validObjClasses,
+              },
+              [head]: {
+                ...descriptionForObjClass(head),
+                selected: true,
```
phorsuedzie: I proposed to put this to the `All` section. There are pro and cons.

Pro:
- Your initially unfolded filter will not hide Objs you wanted to select.
- You do not depend on a "developer's choice".
- Would simplify code ... its only `reduce` then.

Con:
- More Objs displayed immediately.
- Your developer is not able to present you your favoured ObjClass first with a simple reorder of the restriction definition.
- It is easier to simplify the current code than to convert the simple `reduce` into the one with the specific `head`.

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,129 @@
...
+            [head]: {
+              ...descriptionForObjClass(head),
+              selected: true,
             },
+            ...tail.reduce((interimOptions, item) => {
+              return {
+                ...interimOptions,
+                ...{ [item]: descriptionForObjClass(item) },
+              };
+            }, {}),
```
phorsuedzie: FYI: The start value `{}` could be your `[head]` entry.

apepper:
REVIEW (6742894)

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,152 @@
...
-          value: "Image",
+        ...objClasses.reduce((interimOptions, item) => {
+          return {
+            ...interimOptions,
+            ...{ [item]: descriptionForObjClass(item) },
+          };
+        }, {}),
+      },
+    },
+    _modification: MODIFICATION_FILTER,
```
dcsaszar: * [ ] FYI we could prevent some repetition by making the `MODIFICATION_FILTER` include the filter key, and adding it by `...MODIFICATION_FILTER`. After that we could even go one step further and give it a more general name like `MORE_FILTERS` (which is how it appears in the UI).

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,152 @@
...
+        All: {
+          title: "Invalid restrictions",
+          selected: true,
+        },
+      },
+    },
+  };
+}
+
+function objClassFilter(objClass) {
```
dcsaszar: * [ ] FYI We could name this more to the point, e.g. `singleObjClassFilter`. Maybe same for `objClassesFilter` if we can find a better name here.

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,152 @@
...
+
+function descriptionForObjClass(objClass) {
+  const description = DESCRIPTIONS[objClass] || {
+    title: objClass,
+    icon: "question",
+  };
+  return { field: "_objClass", value: objClass, ...description };
+}
+
+const DESCRIPTIONS = {
```
dcsaszar: * [ ] The term "description" is misleading. In the docs only the `title` is a description. Everything else is just called "options". To prevent confusion, we should name it "options", or if it must be more specific something picked from this set: "obj class filter presentation options".

src/config/scrivitoContentBrowser.js
phorsuedzie: I thought this was too ES6, but several lines above the spread is already in use, therefore 👍

src/config/scrivitoContentBrowser.js
phorsuedzie: May I add "label" to the set? :-)

dcsaszar: Whatever 😛. As long we don't contradict https://www.scrivito.com/js-sdk/configureContentBrowser#several%20options

src/Objs/_defaultPageAttributes.js
```diff
@@ -1,6 +1,6 @@
 const defaultPageAttributes = {
   body: ["widgetlist", { only: "SectionWidget" }],
-  navigationBackgroundImage: "reference",
+  navigationBackgroundImage: ["reference", { only: ["Image"] }],
```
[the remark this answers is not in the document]
apepper: Good catch!

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,152 @@
 import * as Scrivito from "scrivito";
 
 Scrivito.configureContentBrowser({
-  filters: {
+  filters: filterContext => {
+    const validObjClasses = filterContext._validObjClasses;
+    if (validObjClasses) {
+      switch (validObjClasses.length) {
+        case 0:
+          return invalidObjClassFilter();
```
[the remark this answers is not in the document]
apepper: One could provide a obj class with a non existing restriction:

```js
poster: ["reference", { only: ["DoesNotExist"] }],
```

In this case `_validObjClasses` will be an empty array. If we use `objClassesFilter([])` in this case, the Content Browser does not render completely:

empty_valid_objs_classes

It also console logs the following error:

```
scrivito_ui.js:168 Error: Missing value (query term: {"field"=>"_obj_class", "operator"=>"equals", "value"=>[]})
```

We could also show `defaultFilter()` if `_validObjClasses.length === 0`.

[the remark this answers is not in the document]
dcsaszar: If this was a common case, raising an error would be the best option, because the original intention can not be fulfilled.

But I don't expect this to happen often, so I can also live with reverting to the default filter.

src/config/scrivitoContentBrowser.js
apepper: How about `filterOptionForObjClass` and `FILTER_OPTIONS`?

apepper:
REVIEW (85b1b80) /cc @dcsaszar @phorsuedzie

src/config/scrivitoContentBrowser.js
```diff
@@ -1,107 +1,152 @@
...
+          field: "_objClass",
+          value: objClasses,
           selected: true,
         },
-        Images: {
-          title: "Images",
-          icon: "image",
-          field: "_objClass",
-          value: "Image",
+        ...objClasses.reduce((interimOptions, item) => {
```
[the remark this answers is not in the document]
dcsaszar: You're right. I wrongly assumed our input was an object. With an array as input we could go for the alternative `zipObject + map`:

```js
zipObject(
  objClasses,
  objClasses.map(descriptionForObjClass)
);
```

src/config/scrivitoContentBrowser.js
dcsaszar: `filterOptionsForObjClass` (+`s`) and `FILTER_OPTIONS` LGTM

src/Objs/_defaultPageAttributes.js
```diff
@@ -1,6 +1,6 @@
 const defaultPageAttributes = {
   body: ["widgetlist", { only: "SectionWidget" }],
-  navigationBackgroundImage: ["reference", { only: ["Image"] }],
+  navigationBackgroundImage: ["reference", { only: ["Image", "Video"] }],
```
dcsaszar: * [ ] The VideoWidget also allows for `Download`. Can we go for consistency?

See https://github.com/Scrivito/scrivito_example_app_js/pull/167/files#diff-6c166dbb21e8fb0e3d946ba385826646R5

src/config/scrivitoContentBrowser.js
```diff
@@ -1,51 +1,36 @@
...
-}
-
-function objClassFilter(objClass) {
+function filterForObjClass(objClass) {
   return {
     _objClass: {
       options: {
         [objClass]: {
-          ...descriptionForObjClass(objClass),
+          ...filterOptionForObjClass(objClass),
```
dcsaszar: * [ ] `filterOptionsForObjClass` is better, because it's at least two options.

src/config/scrivitoContentBrowser.js
```diff
@@ -103,5 +89,42 @@ Scrivito.configureContentBrowser({
         },
       },
     },
-  },
-});
+  };
+}
+
+function filterOptionForObjClass(objClass) {
```
dcsaszar: I just realised this is implemented differently from what I thought when reading the call.

The term filter option becomes very confusing here, because is has multiple meanings. This could be a "filter item"?

The term "description" is still used in a misleading way inside of the function body.

* [ ] Can we try to make this clearer?

apepper:
REVIEW (3914913)

PR #176, opened 2019-03-14
Reduce bundle size by setting webpack node to false

dcsaszar:
I just discovered unexpected polyfill code in my bundle. That's how I found out about this option.
Luckily the Example App had only one dependency on a node feature: `stream` used by `sax` used by `xml`.

### Before/After

image

### DIFF
--- a/src/prerenderContent/prerenderSitemap.js
+++ b/src/prerenderContent/prerenderSitemap.js
@@ -1,6 +1,6 @@
 /* eslint no-console: "off" */
 import * as Scrivito from "scrivito";
-import xml from "xml";
+import xml from "jsontoxml";
 import formatDate from "../utils/formatDate";
 
 export default async function prerenderSitemap(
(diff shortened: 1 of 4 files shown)


src/prerenderContent/prerenderSitemap.js
```diff
@@ -1,6 +1,6 @@
 /* eslint no-console: "off" */
 import * as Scrivito from "scrivito";
-import xml from "xml";
+import xml from "jsontoxml";
```
apepper:
```suggestion
import jsontoxml from "jsontoxml";
```

src/prerenderContent/prerenderSitemap.js
```diff
@@ -25,13 +25,14 @@ function sitemapXml(objClassesWhitelist) {
   const sitemapUrls = pages.map(pageToSitemapUrl);
 
   const content = xml(
```
apepper:
```suggestion
  const content = jsontoxml(
```

apepper: This part is missing and needs to be added, in order to make it work.

apepper: This is the reason, why travis is red.

PR #179, opened 2019-03-26
index.css via css-loader

apepper:
Motivation: This PR sets the base-line to further improve our stylesheets.

Ideally we want a style sheet file per widget/obj, with a base-line `index.scss`. `css-loader` and `MiniCssExtractPlugin` will combine this multiple css files into one big `index.css` file.

As a proof of concept I converted the styling of `TickListWidget`, `VimeoVideoWidget` and `YoutubeVideoWidget` to show, how it is done.

### DIFF
--- a/src/Widgets/TickListWidget/TickListWidgetStyles.scss
+++ b/src/Widgets/TickListWidget/TickListWidgetStyles.scss
@@ -3,19 +3,19 @@
 @import "../../assets/stylesheets/fontawesome/core";
 @import "../../assets/stylesheets/colors";
 
-.TickListWidget__list {
+.tick-list-widget {
   list-style-type: none;
   padding-left: 15px;
   @include clearfix;
 }
-.TickListWidget__list li {
+.tick-list-widget li {
   position: relative;
   padding: 0 0 20px 17px;
   font-family: "PT Sans Narrow", sans-serif;
   font-weight: 400;
   font-size: 20px;
 }
-.TickListWidget__list li:before {
+.tick-list-widget li:before {
   display: block;
   position: absolute;
   top: 2px;
--- a/src/Widgets/TickListWidget/TickListWidgetStyles.scss
+++ b/src/Widgets/TickListWidget/TickListWidgetStyles.scss
@@ -3,19 +3,19 @@
 @import "../../assets/stylesheets/fontawesome/core";
 @import "../../assets/stylesheets/colors";
 
-.tick-list {
+.TickListWidget__list {
   list-style-type: none;
   padding-left: 15px;
   @include clearfix;
 }
-.tick-list li {
+.TickListWidget__list li {
   position: relative;
   padding: 0 0 20px 17px;
   font-family: "PT Sans Narrow", sans-serif;
   font-weight: 400;
   font-size: 20px;
 }
-.tick-list li:before {
+.TickListWidget__list li:before {
   display: block;
   position: absolute;
   top: 2px;
--- a/src/Widgets/TickListWidget/TickListWidgetStyles.scss
+++ b/src/Widgets/TickListWidget/TickListWidgetStyles.scss
@@ -1,3 +1,8 @@
+@import "../../assets/stylesheets/bootstrap/mixins/clearfix";
+@import "../../assets/stylesheets/fontawesome/variables";
+@import "../../assets/stylesheets/fontawesome/core";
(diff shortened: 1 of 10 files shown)


agessler:
👍

src/Widgets/VimeoVideoWidget/VimeoVideoWidgetStyles.scss
```diff
@@ -0,0 +1,7 @@
+.vimeo-fullsize-iframe {
```
dcsaszar: FYI in https://medium.com/seek-blog/block-element-modifying-your-javascript-components-d7f99fcab52b I see a naming convention that uses the component name for CSS namespacing.

* [ ] We should think about a similarly strict naming convention, to prevent misuse of class names.

`VimeoVideoWidgetComponent__fullsize-ifame`, `VimeoVideoWidget__fullsize-ifame`, or so.

src/assets/stylesheets/index.scss
```diff
@@ -122,7 +122,7 @@ html {
 }
 
 body {
-  font-family: "Source Sans Pro", $font-family-sans-serif;
```
dcsaszar: * [ ] Should we also use the simple value in `_variables.scss`, to prevent having two different `sans-serif` families?

src/Widgets/TickListWidget/TickListWidgetStyles.scss
```diff
@@ -1,3 +1,8 @@
+@import "../../assets/stylesheets/bootstrap/mixins/clearfix";
+@import "../../assets/stylesheets/fontawesome/variables";
+@import "../../assets/stylesheets/fontawesome/core";
```
dcsaszar: * [ ] Tha FA core is not widget specific here. Would this belong to a global CSS instead?

apepper: This is needed to make `@extend .fa;` in line 25 work. Otherwise you'll get the following scss error:

```
ModuleBuildError: Module build failed (from ../node_modules/sass-loader/lib/loader.js):

  @extend .fa;
         ^
      ".tick-list li:before" failed to @extend ".fa".
The selector ".fa" was not found.
Use "@extend .fa !optional" if the extend should be able to fail.
```

src/assets/stylesheets/index.scss
apepper: Alex G. and I settled on the simple `font-family: "Source Sans Pro", sans-serif;` version, which does not need an extra dependency.

apepper:
REVIEW (3b0f0f6)

@agessler: Dave suggested to use BEM naming convention for css class names, as described in [Block, Element, Modifying Your JavaScript Components](https://medium.com/seek-blog/block-element-modifying-your-javascript-components-d7f99fcab52b). So every css class should be prefixed with `XWidget__` or `YObj__`.

Here are some examples:
* `fullsize-iframe` becomes `VimeoVideoWidget__fullsize-iframe` or `YoutubeVideoWidget__fullsize-iframe`.
* `tick-list` becomes `TickListWidget__list`.

Are you fine with the new convention?

agessler:
We can do BEM naming convention for css class names if this would be in our DEV sprint plan!

src/assets/stylesheets/index.scss
dcsaszar: To clarify: There's still one use of $font-fmily-sans-serif, which we could *get rid of*. In _varaibles.scss.

dcsaszar:
Talked to @agessler 

BEM seems to be a moving target which also is a bit ugly when mixing camel with dash casing.
So we suggest a different namespace convention:

Use the dashified selector, e.g. `.tick-list-widget`. If modifiers are needed, use a double-dashed prefix: `.vimeo-video-widget--fullsize-iframe`.

WDYT?

apepper:
REVIEW (fede2f6) /cc @agessler @dcsaszar 

Latest css class names:

* `tick-list-widget`
* `vimeo-video-widget--fullsize-iframe`
* `youtube-video-widget--fullsize-iframe`

src/assets/stylesheets/index.scss
apepper: `_varaibles.scss` is part of `bootstrap`. I don't want to modify bootstrap parts.

agessler:
👍

PR #192, opened 2019-06-07
Redirect's openInUi considered harmful

apepper:
Prior to this PR there has been only the `Redirect` obj, that surfed two different purposes: Redirect to a link or open a page in the Scrivito UI.

The second part is a very specific task, that we - more or less - only need on the example content on the first "call to action" button, where novice scrivito editors can try out scrivito. Outside of this narrow use-case no other editor would use it.

Other than that it does not add much value. On the contrary: It can even be harmful, if a regular redirect obj receives the "openInUi" tick. This can lead to a "harmless" redirect to break you page, since regular visitors are also redirected to the Scrivito UI.

This PR adds a dedicated `RedirectToUi` obj, that only has the purpose of redirecting to the UI. The regular `Redirect` obj no longer has the `openInUi` option, to decrease the risk of a redirect gone wrong.

### DIFF
--- a/src/Objs/RedirectToUi/RedirectToUiComponent.js
+++ b/src/Objs/RedirectToUi/RedirectToUiComponent.js
@@ -1,0 +1,37 @@
+import * as React from "react";
+import * as Scrivito from "scrivito";
+import InPlaceEditingPlaceholder from "../../Components/InPlaceEditingPlaceholder";
+
+class RedirectToUiComponent extends React.Component {
+  componentDidMount() {
+    Scrivito.load(() => {
+      const reference = this.props.page.get("reference");
+      const url = reference && Scrivito.urlFor(reference);
+
+      return url;
+    }).then(url => {
+      if (!url) {
+        return;
+      }
+
+      const scrivitoUiUrl = url.replace(/(\/\/[^/]+)/, "$1/scrivito");
+      window.top.location.replace(scrivitoUiUrl);
+    });
+  }
+
+  render() {
+    const reference = this.props.page.get("reference");
+
+    if (!reference) {
+      return (
+        <InPlaceEditingPlaceholder center>
+          Select a reference in the redirect to ui page properties.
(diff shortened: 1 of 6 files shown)


src/Objs/RedirectToUi/RedirectToUiComponent.js
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+import * as Scrivito from "scrivito";
+
+class RedirectToUiComponent extends React.Component {
+  componentDidMount() {
+    const scrivitoUiUrl = window.location.origin + "/scrivito/";
```
dcsaszar: * [ ] FYI `origin` is IE11+ only https://developer.mozilla.org/en-US/docs/Web/API/Window/location

src/Objs/RedirectToUi/RedirectToUiEditingConfig.js
```diff
@@ -0,0 +1,8 @@
+import * as Scrivito from "scrivito";
+import redirectObjIcon from "../../assets/images/redirect_obj.svg";
+
+Scrivito.provideEditingConfig("RedirectToUi", {
+  title: "Redirect To Ui",
```
dcsaszar: * [ ] `UI` - upper case.

src/Objs/RedirectToUi/RedirectToUiComponent.js
apepper: Our UI does not even support IE11. To quote from https://www.scrivito.com/user-guide:

> Supported browsers for editing
> * Google Chrome: latest 2 versions
> * Mozilla Firefox: latest 2 versions
> * Apple Safari: latest version
> * Microsoft Edge: latest version

dcsaszar: I see, didn't have in mind that this is a UI feature of the app. 😝

PR #195, opened 2019-06-25
Fixed styles of feature panel widget

karminkarmen:
This PR fixes CSS bug in `feature panel` widget. 

before fix:
Screenshot 2019-06-25 at 12 58 35

after fix:
Screenshot 2019-06-25 at 12 59 21


apepper:
Thank you for your effort! I did take a closer look, why this even broke and found #179 to be the source. I did build a PR in #196 that fixes the issue in another way.

PR #196, opened 2019-06-25
Inlined TickListWidgetStyles.scss again.

apepper:
Otherwise we run into problems with font awesome variables breaking the style of the Feature Panel.


src/assets/stylesheets/index.scss
```diff
@@ -3114,6 +3114,31 @@ img.cookie-img {
   }
 }
 
+/* list *
```
dcsaszar: * [ ] FMI Was this the original comment? Looks a bit sparse. 😛

src/assets/stylesheets/index.scss
```diff
@@ -3114,6 +3114,31 @@ img.cookie-img {
...
+.tick-list-widget li:before {
+  display: block;
+  position: absolute;
+  top: 2px;
+  left: -15px;
+  content: "\f00c" !important;
+  color: $secondary;
+  @extend .fa;
+}
+
```
dcsaszar: * [ ] FYI Not related to this PR, bit the following line looks "interesting"

src/assets/stylesheets/index.scss
apepper: Both comments are by our css designer.

PR #200, opened 2019-07-30
[js] Use ReactDOMServer.renderToString and ReactDOM.hydrate.

apepper:
Motivation: less flicker during "reanimate".

Makes use of the newly added `Scrivito.isEditorLoggedIn()`.


apepper:
REVIEW (6b312fa)

PR #201, opened 2019-07-31
Update cookie banner to be GDPR compliant

X3phiroth:
To be GDPR compliant this adds the option to deny permission to use GoogleAnalytics by declining third party cookies in the cookie banner.


src/Components/CookieConsent.js
```diff
@@ -22,10 +22,13 @@ function CookieConsent() {
   return (
     <ReactCookieConsent
       buttonText="accept"
+      enableDeclineButton
+      declineButtonText="decline"
       disableStyles
       containerClasses="cookie-box d-lg-flex d-sm-flex align-items-center"
       contentClasses="cookie-content d-flex align-items-center"
       buttonClasses="cookie-button btn btn-primary ml-auto"
+      declineButtonClasses="cookie-button btn btn-secondary ml-auto mr-2"
```
apepper: `btn-secondary` is the correct one here, but I noticed, that it does not look optimal.

Probably during the boostrap 3 to 4 conversion something went wrong with `btn-secondary`. The color of it should be `#fff`. So simply add `.btn-secondary` right along `.btn-primary` and `btn-success` in `index.scss` line number `392`.

src/Components/CookieConsent.js
```diff
@@ -22,10 +22,13 @@ function CookieConsent() {
   return (
     <ReactCookieConsent
       buttonText="accept"
+      enableDeclineButton
+      declineButtonText="decline"
```
apepper: I would also use the option `flipButtons` to make the accept button before the decline button. Than `mr-2` needs to move from `declineButtonClasses` to `buttonClasses`.

src/Components/GoogleAnalytics.js
```diff
@@ -37,7 +38,10 @@ class GoogleAnalytics extends React.Component {
   }
 
   render() {
-    if (!this.state.trackingId) {
+    if (
+      !this.state.trackingId ||
+      !(getCookieValue("CookieConsent") === "true")
```
apepper: Let's create a more high level method for the cookie consent. How about `cookieConsentGiven()`?

It could looks something like this (not yet tested):

```js
import { Cookies } from "react-cookie-consent";

export default function cookieConsentGiven() {
  return Cookies.get("CookieConsent") === "true";
}
```

If so we can simply remove the `getCookieValue` method.

src/Components/GoogleAnalytics.js
```diff
@@ -1,6 +1,7 @@
 import * as React from "react";
 import * as Scrivito from "scrivito";
 import Helmet from "react-helmet";
+import * as cookieConsentGiven from "../utils/cookieConsentGiven";
```
dcsaszar: * [ ] Please use the non-namespace import.

PR #204, opened 2019-08-01
added eslint-config-react-app

7ack7:
added eslint-config-react-app and fixed related warnings


src/Widgets/TableWidget/TableWidgetComponent.js
```diff
@@ -69,8 +69,7 @@ const AddMoreRows = Scrivito.connect(
       <tr>
         <th colSpan="4">
           <div className="text-center">
-            <a
-              href="#"
+            <span
```
apepper: Now the "Click to add new row" does not indicate anymore, that it can be clicked (e.g. because the hover state is missing).

src/Widgets/TestimonialSliderWidget/TestimonialSliderWidgetComponent.js
```diff
@@ -61,7 +63,7 @@ function sliderSettings(testimonials) {
     dotsClass: "quote-portrait-wrapper",
     customPaging: i => {
       const image = testimonialAuthorImages[i];
-      return <a>{image}</a>;
+      return <span>{image}</span>;
```
apepper: The testimonial slider now looks simply broken.

src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
```diff
@@ -139,17 +139,13 @@ const Thumbnail = Scrivito.connect(({ widget, openLightbox, currentTag }) => {
         className="gallery-box-image"
         style={{ background: { image } }}
       />
-      <a
-        href="#"
-        className="gallery-box-content-wrapper"
-        onClick={openLightbox}
-      >
+      <span className="gallery-box-content-wrapper" onClick={openLightbox}>
```
apepper: There is now no indication, that this can be clicked (e.g. no hover state)

7ack7:
@apepper done

PR #209, opened 2019-08-07
Style/chunk split

7ack7:
- moved variables to another folder
- chunked TickListWidget style


src/assets/stylesheets/variables/_layout.scss
```diff
@@ -0,0 +1,6 @@
+$grid-columns: 12;
+$grid-gutter-width: 30px;
```
apepper: Both of these values are boostrap default values. I guess it's save to remove these grid variables.

src/assets/stylesheets/variables/_layout.scss
```diff
@@ -0,0 +1,6 @@
...
+
+// use higher zindex, to avoid edit marker to be shown
+$zindex-modal-backdrop: 100040 !default;
+$zindex-modal: 100050 !default;
```
apepper: I looked hard, but if I'm not mistaken, the `$zindex-modal-backdrop` and `$zindex-modal` are only used for the `modal` css class by Bootstrap. If I'm not mistaken this css class is never used.

So I guess we can actually get rid of all variables in `_layout.scss`. What do you think?

src/assets/stylesheets/variables/_responsive.scss
```diff
@@ -0,0 +1,6 @@
+$screen-xs-max: 767px;
```
apepper: Please add the following comment on the very top of the file:

```scss
/* - BS3 variables - */
```

This will help our designer to identify the source of these variables.

src/assets/stylesheets/variables/_layout.scss
7ack7: Sure, I will just remove it then.

src/Widgets/TickListWidget/tickListWidget.scss
```diff
@@ -0,0 +1,27 @@
+@import "src/assets/stylesheets/bootstrap/mixins/clearfix";
+@import "src/assets/stylesheets/fontawesome/variables";
+@import "src/assets/stylesheets/fontawesome/core";
```
apepper:
```suggestion
@import "src/assets/stylesheets/fontawesome/mixins";
```

src/Widgets/TickListWidget/tickListWidget.scss
```diff
@@ -0,0 +1,27 @@
...
+  font-size: 20px;
+
+  &:before {
+    display: block;
+    position: absolute;
+    top: 2px;
+    left: -15px;
+    content: "\f00c" !important;
+    color: $secondary;
+    @extend .fa;
```
apepper:
```suggestion
    @include fa-icon;
```

`@extend .fa` does actually redefine `.fa` globally. So then the order of which the widgets css and the index.css are loaded becomes important. Ideally the order should not matter.

So if we use a mixin instead, `.fa` does not get redefined globally.

7ack7:
@apepper done

apepper:
@Agsiegert Heads up: We changed the location of `_colors.scss`. If I'm not mistaken, they are referenced in https://www.scrivito.com/changing-colors-in-the-example-app-5ec76614496973dd and https://www.scrivito.com/choosing-a-color-theme-in-the-scrivito-example-app-e88bfce096846fc2.

PR #211, opened 2019-08-08
Style/split chunk

7ack7:
- chunked thumbnailGalleryWidget styles

- moved mixins to another file

- BEM naming convention

### DIFF
--- a/src/Widgets/ThumbnailGalleryWidget/thumbnailGalleryWidget.scss
+++ b/src/Widgets/ThumbnailGalleryWidget/thumbnailGalleryWidget.scss
@@ -1,11 +1,13 @@
 @import "src/assets/stylesheets/variables";
 @import "src/assets/stylesheets/mixins";
-@import "src/assets/stylesheets/bootstrap/bootstrap";
+@import "src/assets/stylesheets/bootstrap/functions";
+@import "src/assets/stylesheets/bootstrap/variables";
+@import "src/assets/stylesheets/bootstrap/mixins/transition";
 
-.gutter0 .gallery-box__wrapper {
+.gutter0 .thumbnail-gallery-widget--wrapper {
   padding: 0;
 }
-.gallery-box {
+.thumbnail-gallery-widget {
   overflow: hidden;
   max-height: 800px;
   max-width: 800px;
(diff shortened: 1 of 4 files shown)


src/Widgets/ThumbnailGalleryWidget/thumbnailGalleryWidget.scss
```diff
@@ -0,0 +1,120 @@
+@import "src/assets/stylesheets/variables";
+@import "src/assets/stylesheets/mixins";
+@import "src/assets/stylesheets/bootstrap/bootstrap";
```
apepper: Adding the complete `bootstrap` to a widget's css is not a good idea. This instructs sass and webpack to add a full copy of bootstrap to the final `index.css`. With this is actually got really big (see https://travis-ci.org/Scrivito/scrivito_example_app_js/jobs/569311355#L304) compared to the current master (see https://travis-ci.org/Scrivito/scrivito_example_app_js/jobs/569301877#L294). So increasing the `index.css` by over 50% (more than 100 kb) is not an option.

If you do make use of bootstrap, please only import it partially and ideally only mixins (e.g. `@import "src/assets/stylesheets/bootstrap/mixins/clearfix";`).

Also: Whenever you're using a `@extend` directive in a widget's scss file please try to replace it with a `@include` rule or a plain css style. Otherwise we run into the risk of overwriting the original `@extend` method, as previously experienced with the `@extend fa` in a previous PR.

src/Widgets/ThumbnailGalleryWidget/thumbnailGalleryWidget.scss
```diff
@@ -0,0 +1,120 @@
...
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    /*  display: flex;
+    justify-content: center;
+    align-items: center; */
+    @extend .d-flex;
+    @extend .align-items-center;
+    @extend .justify-content-center;
```
apepper: Please replace `@extend` rules with mixings (via `@include`)  or a plain css directive. See motivation for this in the bootstrap section above.

7ack7:
@apepper done

PR #212, opened 2019-08-12
chunked testimonialSliderWidget styles

7ack7:
@apepper

### DIFF
--- a/src/Widgets/TestimonialSliderWidget/testimonialSliderWidget.scss
+++ b/src/Widgets/TestimonialSliderWidget/testimonialSliderWidget.scss
@@ -3,43 +3,45 @@
 @import "src/assets/stylesheets/bootstrap/variables";
 @import "src/assets/stylesheets/bootstrap/mixins/transition";
 
-.quote-headline {
-  color: $secondary !important;
-  font-size: 200px;
-  line-height: 0;
-  margin: 65px 0 0 0;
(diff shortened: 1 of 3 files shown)


src/Widgets/TestimonialSliderWidget/testimonialSliderWidget.scss
```diff
@@ -0,0 +1,45 @@
+@import "src/assets/stylesheets/variables";
+@import "src/assets/stylesheets/bootstrap/functions";
+@import "src/assets/stylesheets/bootstrap/variables";
+@import "src/assets/stylesheets/bootstrap/mixins/transition";
+
+.quote-headline {
```
apepper: Please change the css namespace to the widget name and its function. So in this case:

```suggestion
.testimonial-slider-widget {
  &--quote-headline {
```

src/Widgets/TestimonialSliderWidget/testimonialSliderWidget.scss
```diff
@@ -0,0 +1,45 @@
...
+  color: $secondary !important;
+  font-size: 200px;
+  line-height: 0;
+  margin: 65px 0 0 0;
+  vertical-align: middle;
+}
+
+.quote-portrait-wrapper {
```
apepper: Here the same as above.

PR #213, opened 2019-08-12
Chunk/table widget

7ack7:
I noticed that when icon is placed on the footer it is the same color as the background so I changed it to white.

Screenshot 2019-08-12 at 13 04 06

@apepper


src/Widgets/TableWidget/tableWidget.scss
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    th,
+    td {
+      background: $secondary;
+      color: #fff;
+    }
+  }
+
+  .quantity {
```
apepper: This is actually styling for the `PriceWidget`. Could you extract this `qunatitiy` part to a `PriceWidget.scss`, which is used by the `PriceWidget` in this PR as well? Currently `PriceWidget` can only be used in a `TableWidget`, but I don't want the code to be too much coupled.

src/assets/stylesheets/index.scss
```diff
@@ -617,7 +617,7 @@ $easing: cubic-bezier(0.7, 0, 0, 0.7);
   .h4,
   .h5,
   .h6,
-  .table-features,
+  .table-widget,
```
apepper: This rings an 🚨 for me. Ideally all styling needed for a widget should be in its corresponding `.scss` file. But this here indicates, that their is some global styling to be also considered.

I think it is better, if we duplicate the `color: #fff` part for `.bg-dark-image` in `TableWidget.scss`.

This also applies to the `.footer` section.

7ack7:
@apepper done

src/Widgets/TableWidget/TableWidget.scss
```diff
@@ -0,0 +1,184 @@
+@import "src/assets/stylesheets/variables";
+
+.bg-dark-image,
+.bg-brand-primary,
+.bg-brand-secondary,
+.bg-grey,
+.bg-greydark,
+.bg-greymiddle,
+.bg-greymiddle {
+  footer & {
```
apepper: If I'm not mistaken, the `footer` part can be removed, since the general `.table-widget` overwrite is strong enough anyway.

7ack7:
@apepper I missed that conflict, now should be fine;)

PR #214, opened 2019-08-12
chunked pricingWidget styles

7ack7:


src/Widgets/PricingWidget/PricingWidgetComponent.js
```diff
@@ -1,11 +1,16 @@
 import * as React from "react";
 import * as Scrivito from "scrivito";
 
+import "./pricingWidget.scss";
```
apepper: Please use `CamelCase.scss` here (and of cause change the name of the original file)

```suggestion
import "./PricingWidget.scss";
```

7ack7:
@apepper done

PR #220, opened 2019-08-19
chunked LinkContainerWidget styles

7ack7:
- chunked LinkContainerWidget styles
- added b-bottom class in JSX instead of extending `.link-container-widget--headline` with it

### DIFF
--- a/src/Widgets/LinkContainerWidget/LinkContainerWidget.scss
+++ b/src/Widgets/LinkContainerWidget/LinkContainerWidget.scss
@@ -3,22 +3,13 @@
 .link-container-widget {
   list-style-type: none;
   padding: 0;
-  display: block;
-  color: $theme-greydark;
 
-  li {
+  &,
+  & li,
+  & li a,
+  & li span {
     display: block;
     color: $theme-greydark;
-
-    a {
-      display: block;
-      color: $theme-greydark;
-    }
-
-    span {
-      display: block;
-      color: $theme-greydark;
-    }
   }
 
   .b-bottom {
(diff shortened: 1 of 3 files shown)


src/Widgets/LinkContainerWidget/LinkContainerWidget.scss
```diff
@@ -0,0 +1,74 @@
...
+    color: $theme-greydark;
+
+    a {
+      display: block;
+      color: $theme-greydark;
+    }
+
+    span {
+      display: block;
+      color: $theme-greydark;
```
apepper: I would move this part out of `.link-container-widget` and instead write this:

```scss
.link-container-widget,
.link-container-widget li,
.link-container-widget li a,
.link-container-widget li span {
  display: block;
  color: $theme-greydark;
}
```

src/Widgets/LinkContainerWidget/LinkContainerWidget.scss
```diff
@@ -0,0 +1,74 @@
...
+    text-decoration: none;
+
+    &:hover {
+      text-decoration: none;
+    }
+
+    &:focus {
+      text-decoration: none;
+    }
+  }
```
apepper: This can be simplified:

```scss
  a,
  a:hover,
  a:focus {
    text-decoration: none;
  }
```

src/Widgets/LinkContainerWidget/LinkContainerWidget.scss
7ack7: or maybe like this inside `.link-container-widget`?

```suggestion
  &,
  & li,
  & li a,
  & li span {
    display: block;
    color: $theme-greydark;
  }
```

apepper: If that works, great!

apepper: If that works, great!

7ack7:
@apepper done

PR #221, opened 2019-08-19
chunked GalleryWidget styles

7ack7:


src/Widgets/GalleryWidget/GalleryWidgetComponent.js
```diff
@@ -16,8 +18,8 @@ Scrivito.provideComponent("GalleryWidget", ({ widget }) => {
 
   const settings = sliderSettings(images);
   return (
-    <div className="slick-gallary-fluid">
-      <Slider {...settings} className="slickslide">
+    <div className="gallery-widget">
+      <Slider {...settings} className="gallery-widget--slick-slide">
```
apepper: Their is no `gallery-widget--slick-slide` defined. But a `slick-slider` under `gallery-widget`.

Since the default class of a slick slider is `slick-slider` I would simply remove the custom `className`:

```suggestion
      <Slider {...settings}>
```

Please not that `slickslide` and `slick-slider` in the previous version where added to the exact same dom element. So consolidating the two into `.galery-widget .slick-slider` seems to be a good idea.

src/Widgets/GalleryWidget/GalleryWidget.scss
```diff
@@ -0,0 +1,142 @@
...
+    margin: 0;
+    padding: 0;
+    position: relative;
+
+    .slick-track {
+      min-width: 100%;
+    }
+  }
+
+  .slick-slide {
```
apepper: See my other comment regarding `slickslide` and `slick-slider`:

```suggestion
  .slick-slider {
```

This can then be of cause be combined with the upper `.slick-slider`.

src/Widgets/GalleryWidget/GalleryWidget.scss
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    li,
+    button:before {
+      margin: 0;
+      padding: 0;
+    }
+  }
+
+  .slick-slide {
```
apepper: See my other comment regarding `slickslide` and `slick-slider`:

```suggestion
  .slick-slider {
```

This can then be of cause be combined with the upper `.slick-slider`.

7ack7:
@apepper done

PR #223, opened 2019-08-20
chunked FactWidget styles

7ack7:


src/Widgets/FactWidget/FactWidget.scss
```diff
@@ -0,0 +1,45 @@
...
+.bg-greydark,
+.bg-dark-image {
+  .fact-widget .key {
+    color: $theme-greylight;
+  }
+}
+
+.bg-brand-primary .fact-widget .value,
+.bg-brand-secondary .fact-widget .value {
+  color: #fff;
```
apepper: How about this?

```scss
.bg-brand-primary,
.bg-brand-secondary {
  .fact-widget .value {
    color: #fff;
  }
}
```

7ack7: right ;p

PR #226, opened 2019-08-21
chunked CarouselWidget styles

7ack7:


7ack7:
@apepper done

src/Widgets/CarouselWidget/CarouselWidget.scss
```diff
@@ -0,0 +1,152 @@
...
+  }
+}
+
+@media (max-width: $screen-xs-max) {
+  .carousel-widget {
+    .carousel-images .carousel-item {
+      height: 250px;
+    }
+
+    .description-box {
```
apepper: Since we are now more strict with where what media query applies, the `screen-xs-max` part is missing the following that it previously "inherited" by `max-width: $screen-sm-max`:

```suggestion
    .description-box {
      display: block;
      
```

src/Widgets/PricingWidget/PricingWidget.scss
```diff
@@ -121,7 +121,7 @@
   }
```
apepper: Here we also need to watch out. Previously less than 600px used the style of `$screen-sm-max` and only overwrote `widght: 90% !important`. Since the `$screen-sm-max` style is now independent we need to add the missing styles to `600px`:

```suggestion
@media (max-width: 600px) {
  .pricing-widget {
    .pricing-widget-chart {
      float: none !important;
      left: 0px !important;
      top: 0px !important;
      margin: 0 auto;
      width: 90% !important;
      margin-bottom: 25px !important;

      &.last {
        margin-bottom: 0px !important;
      }
    }
  }
}
```

7ack7:
sorry @apepper, should be good now

PR #229, opened 2019-08-23
Chunk/navbar

7ack7:
- Chunked Navigation styles
- Moved import of App.js  under the index.scss because of conflicting bootstrap styles


src/Components/Navigation/Navigation.scss
```diff
@@ -0,0 +1,695 @@
...
+    &.open {
+      > a {
+        background: $nav-hover !important;
+        color: #fff !important;
+      }
+
+      > .menu-toggle {
+        background: $nav-hover !important;
+        color: #fff !important;
+      }
```
apepper: Would the following work?

```scss
    &.open {
      > a,
      > .menu-toggle {
        background: $nav-hover !important;
        color: #fff !important;
      }
    }
```

src/assets/stylesheets/index.scss
```diff
@@ -673,618 +673,6 @@ section.navbar-fixed {
...
-.nav.navbar-right {
-  margin-left: auto;
-  margin-right: 45px;
-}
-
-.nav > li,
-.nav > li > a {
-  position: relative;
-  display: block;
-}
```
apepper: If I'm not mistaken the following part got lost:

```scss
.nav > li > a {	
  position: relative;	
  display: block;	
}
```

src/Components/Navigation/Navigation.scss
```diff
@@ -0,0 +1,695 @@
...
+
+.nav {
+  margin: 0;
+
+  &.navbar-right {
+    margin-left: auto;
+    margin-right: 45px;
+  }
+
+  > li {
```
apepper: If I'm not mistaken the following is missing (see my previous comment):

```suggestion
  > li,
  > li > a {
```

src/Components/Navigation/Navigation.scss
```diff
@@ -0,0 +1,695 @@
...
+      > .menu-toggle {
+        background: $nav-hover !important;
+        color: #fff !important;
+      }
+    }
+
+    &.active > .menu-toggle {
+      background: $primary !important;
+      color: #fff !important;
+    }
```
apepper: I think `a` is missing here:


```scss
    &.active {
      > a,
      > .menu-toggle {
        background: $primary !important;
        color: #fff !important;
      }
    }
```

src/assets/stylesheets/index.scss
```diff
@@ -673,618 +673,6 @@ section.navbar-fixed {
...
-  background: $nav-hover;
-  color: #fff;
-}
-.nav .dropdown-menu > li.active > a:hover,
-.nav .dropdown-menu > li.active > a:focus {
-  background: $primary;
-  color: #fff;
-}
-.open > .dropdown-menu {
-  display: block;
```
apepper: Where did this rule go? Is it not important?

src/Components/Navigation/Navigation.scss
```diff
@@ -0,0 +1,695 @@
...
+
+    .icon-bar {
+      display: block;
+      height: 2px;
+      border-radius: 1px;
+      background: $secondary;
+      width: 30px;
+
+      + .icon-bar {
```
apepper: For my information: What does the `+` syntax do?

src/assets/stylesheets/index.scss
```diff
@@ -673,618 +673,6 @@ section.navbar-fixed {
...
-  padding: 13px 20px 10px;
-}
-.nav-centered .input-group .btn .fa {
-  font-size: 22px;
-}
-@media (max-width: $screen-xs-max) {
-  .nav-centered .input-group .btn .fa {
-    padding-left: 0;
-  }
-}
```
apepper: If I'm not mistaken, the whole `input-group` section got lost.

src/index.js
```diff
@@ -4,9 +4,9 @@ import * as ReactDOM from "react-dom";
 import * as Scrivito from "scrivito";
 import "./Objs";
 import "./Widgets";
+import "./assets/stylesheets/index.scss";
 import App from "./App";
 import "./config";
-import "./assets/stylesheets/index.scss";
```
apepper: Why is this change needed? Which styling in `index.scss` overwrites styles of `Navigation.scss`?

My goal would be to have `index.scss` loaded _last_. We either have to fix the stylings in `Novigation.scss` accordingly or import `Navigation.scss` via `index.scss` and not via the js component.

src/assets/stylesheets/index.scss
apepper: It's here: https://github.com/Scrivito/scrivito_example_app_js/pull/229/files#diff-9e04cc0a7b0d44695b075612dc170e51R283

src/Components/Navigation/Navigation.scss
7ack7: Selects all .icon-bar elements that are placed immediately after .icon-bar elements in this case.

src/index.js
7ack7: `.menu-toggle .fa` & `.navbar-brand` at least 

Screenshot 2019-09-04 at 13 46 41
Screenshot 2019-09-04 at 13 47 01

so maybe I can add it to imports in index.scss as you suggested?

apepper: If you don't find another way simply import it via `index.scss`.

src/assets/stylesheets/index.scss
7ack7: I think this part is not needed, it only breaks nav items which have arrows but without it everything looks fine for me when files are imported in the proper order. Hopely I am not missing something here.

apepper:
Continued as https://github.com/Scrivito/scrivito_example_app_js/pull/238.

PR #230, opened 2019-08-23
Replaced YoutubeVideoWidget with external npm package.

apepper:


apepper:
Closing for now, since the external package adds more payload than needed.

PR #236, opened 2019-09-04
Removed unused styles

7ack7:
Additionaly I removed `.margin-neg-offset` because I couldn't find it used anywhere neither

### DIFF
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -1734,21 +1734,6 @@ _:-ms-fullscreen,
   }
 }
 
-/* gallery thumbs
-=================================================================== */
-
-.gallery-thumbs img {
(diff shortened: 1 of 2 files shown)


src/assets/stylesheets/index.scss
```diff
@@ -2107,67 +2075,6 @@ img.cookie-img {
...
-.floating-image.width-100 {
-  img {
-    width: 100%;
-  }
-}
-
-/* google maps fix
-================================================== */
-
-.gm-style img {
```
apepper: Actually `gm-style` is used indirectly in the `GoogleMapsWidget`, if you set it to `Interactive` (Widget property).

Can you move this section to `GoogleMapsWidget.scss`?

7ack7: oh oki

PR #239, opened 2019-09-09
Cookie banner improvement

X3phiroth:
Add functionality to bind GA directly on accept 3rd party cookies and not just after the next refresh.


src/utils/cookieConsentGiven.js
```diff
@@ -1,5 +1,23 @@
 import { Cookies } from "react-cookie-consent";
+import * as Scrivito from "scrivito";
+
+let resolveCookie;
+const cookiePromise = new Promise(resolve => (resolveCookie = resolve));
+
+const interval = setInterval(() => {
```
apepper: From an architectural and performance standpoint: Polling should only be the last resort, if no other measures can be used.

I suggest a different workflow:

* You already have a the `resolveCookieConsent` method exported.
* Use the `resolveCookieConsent` in `CookieConsent` once a click on accept occures.
* Add a `componentDidMount` in `CookieConsent`, that calls `resolveCookieConsent` is the cookie is set (only one time).

This approach has different advantages:
1. There is no polling. The promise simply waits until either `componentDidMount` or `onAccept` of `CookieConsent` took place.
2. The implementation of how the consent is given and stored is then encapsulated in `CookieConsent`.

PR #240, opened 2019-09-09
chunked BlogPost timeline styles

apepper:
Continues #228.


apepper:
LGTM (a953a4d)

Good job @7ack7 for the initial work!

PR #241, opened 2019-09-09
[generator-scrivito] Also generate a scss file for new objs/widgets.

apepper:


apepper:
REVIEW (02f22a7)

Good idea. I went with option 2.

PR #247, opened 2019-09-30
Use privacy enhanced YouTube embed

dcsaszar:
See https://www.ghacks.net/2018/05/23/why-you-should-always-use-youtubes-privacy-enhanced-mode/ for background


dcsaszar:
#248

PR #251, opened 2019-10-07
Eslint 5 => 6

apepper:


.eslintrc
```diff
@@ -22,6 +22,12 @@
     "btoa": true
   },
   "rules": {
+    "class-methods-use-this": [
```
dcsaszar: * [ ] How come we're using `eslint-plugin-react` and still need a rule like this? We should check the config.

apepper:
REVIEW (d32e838)

PR #261, opened 2019-11-14
Replaced ExtendCspHeadersWebpackPlugin by much simpler transform action

apepper:


apepper:
REVIEW (6420278)

Simplified the code even further.

PR #270, opened 2020-01-16
404 bugfix

apepper:
This PR does two things:

* Do not redirect on the 404 page, if in the Scrivito UI.
* Use a query param (a.k.a. "search") instead of the path for the 404 status code indicator. Also leave all other parts of the URL as is.

### DIFF
--- a/src/Components/NotFoundErrorPage.js
+++ b/src/Components/NotFoundErrorPage.js
@@ -8,9 +8,7 @@ class NotFoundErrorPage extends React.Component {
     const { search } = window.location;
 
     if (!Scrivito.isEditorLoggedIn() && !search.includes(statusCode)) {
-      window.location.search = search
-        ? `${search}&${statusCode}`
-        : `?${statusCode}`;
+      window.location.search = search ? `${search}&${statusCode}` : statusCode;
     }
   }
 
--- a/src/Components/NotFoundErrorPage.js
+++ b/src/Components/NotFoundErrorPage.js
@@ -4,10 +4,13 @@ import { Helmet } from "react-helmet-async";
 
 class NotFoundErrorPage extends React.Component {
   componentDidMount() {
-    const path = window.location.pathname;
+    const statusCode = "status-code=404-not-found";
+    const { search } = window.location;
 
-    if (!Scrivito.isEditorLoggedIn() && path !== "/404") {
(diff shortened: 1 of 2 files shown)


src/Components/NotFoundErrorPage.js
```diff
@@ -4,10 +4,13 @@ import { Helmet } from "react-helmet-async";
 
 class NotFoundErrorPage extends React.Component {
   componentDidMount() {
-    const path = window.location.pathname;
+    const statusCode = "status-code=404-not-found";
+    const { search } = window.location;
 
-    if (path !== "/404") {
-      window.location.replace(`/404?path=${encodeURIComponent(path)}`);
+    if (!Scrivito.isEditorLoggedIn() && !search.includes(statusCode)) {
```
dcsaszar: * [ ] FYI if we make the `isEditorLoggedIn` a more early guard, and/or the other check a guard, the readability may improve - but I haven't tried, so I may be wrong.

src/Components/NotFoundErrorPage.js
```diff
@@ -4,10 +4,13 @@ import { Helmet } from "react-helmet-async";
...
+      window.location.search = search
+        ? `${search}&${statusCode}`
+        : `?${statusCode}`;
```
dcsaszar: Per spec we don't need to add the `?` here. https://html.spec.whatwg.org/multipage/links.html#dom-hyperlink-search
* [ ] FYI can be simplified

src/Components/NotFoundErrorPage.js
apepper: Is that really better?

```js
    if (Scrivito.isEditorLoggedIn()) {
      return;
    }
    const statusCode = "status-code=404-not-found";
    const { search } = window.location;
    if (search.includes(statusCode)) {
      return;
    }

    window.location.search = search
      ? `${search}&${statusCode}`
      : `?${statusCode}`;
```

One line if/return statements are currently not allowed by the eslint rules.

apepper:
REVIEW (4269b72)

src/Components/NotFoundErrorPage.js
dcsaszar: No, it isn't. The latest version of `componentDidMount` became more readable already.

* [ ] FYI One more idea is making the `const statusCode = "status-code=404-not-found";`  an actual toplevel const. No need to put it in the function body.

PR #278, opened 2020-02-04
update bootstrap 4.0 -> 4.4.1

agessler:


apepper:
Thank you for your work! It looks good and works fine.

But the resulting CSS file (`index.css`) grew by 10% (from 221 to 244 KB). This file is critical for lighthouse performance, so a 10% raise is not a good idea.

Since we use none of the new 4.4 features I think we should simply stick to version 4.0 for now.

PR #281, opened 2020-02-05
Design: Search suggest

agessler:
@apepper HTML here https://github.com/infopark/scrivito_example_app_js_content/pull/39/commits/8dfa9c7611ac1883f0a4d114316641be88e17514


src/assets/stylesheets/index.scss
```diff
@@ -1170,6 +1170,37 @@ img.cookie-img {
   }
 }
 
+/* search-suggest
+=================================================================== */
+
+.search-box-suggest {
```
apepper: Can you move this part to https://github.com/Scrivito/scrivito_example_app_js/blob/master/src/assets/stylesheets/_navigation.scss#L336, since we want to make the css more modular?

src/assets/stylesheets/index.scss
```diff
@@ -1170,6 +1170,37 @@ img.cookie-img {
...
+.search-box-suggest li:first-child { box-shadow: none;}
+.search-box-suggest li.active,
+.search-box-suggest li:hover { background: #fff; cursor: pointer;}
+
+.search-box .search-box-suggest {
+  width: calc(100% - 40px); position: absolute;
+  top: 100%; left: 0;
+  }
+
+.nav-centered .search-box-suggest {
```
apepper: Same here: https://github.com/Scrivito/scrivito_example_app_js/blob/master/src/assets/stylesheets/_navigation.scss#L590

src/assets/stylesheets/index.scss
agessler: You can cut it out and place it in a separate SCSS file. But this does not mean that you can just copy an paste it .in a different project. It depends to this site design. For me it does not make sense.

src/assets/stylesheets/index.scss
agessler: same here.

agessler:
@apepper new commit

PR #282, opened 2020-02-05
Fix prerendering time difference in blog post list

dcsaszar:
The time difference on the rendering server and client lead to subtle issues in the list of blog posts.

For instance: `<time class="timeline-badge" dateTime="2019-09-02T22:00:00.000Z"><div>03/09</div></time>` vs. `<time class="timeline-badge" dateTime="2019-09-02T22:00:00.000Z"><div>02/09</div></time>`.

The fix is according to https://stackoverflow.com/a/50884055.


dcsaszar:
Closing for now. We realized that these time differences were not the cause of our issue. We'll keep looking if there are actual issues regarding date handling.

PR #283, opened 2020-02-06
Fingerprint CSS files

apepper:
Adds fingerprints to css files. This simplifies caching on other hosters.


src/prerenderContent/generateHtml.js
```diff
@@ -1,11 +1,17 @@
...
+
+export default async function generateHtml({
   objId,
   htmlAttributes,
   headContent,
   bodyAttributes,
   bodyContent,
   preloadDumpFileName,
 }) {
+  const assets = await assetManifest;
```
dcsaszar: * [ ] FYI This code may be a bit more complicated than it needs to be. We're using a top-level const only once, so inlining would be good enough. Also we're mixing `then` and `await`.

webpack.config.js
```diff
@@ -166,11 +168,25 @@ function generatePlugins({ isProduction, isPrerendering, scrivitoOrigin }) {
       },
     ]),
     new MiniCssExtractPlugin({
-      filename: "[name].css",
+      filename: "assets/[name].[contenthash].css",
+    }),
+    new HtmlWebpackPlugin({
+      filename: "catch_all_index.html",
+      template: "catch_all_index.html",
+      inject: false,
```
dcsaszar: * [ ] Our source files have self-closing `link` tags. According to the docs,  `xhtml: true` renders the link tags as self-closing (XHTML compliant).
I don't know it that only holds true for injected link tags, or all link tags, but for consistency we should turn it on.

src/prerenderContent/generateHtml.js
apepper: > Also we're mixing then and await.

If I rewrite the code to this:

```ts
const assetManifestResponse = await fetch("asset-manifest.json");
const assetManifest = await assetManifestResponse.json();
```

Then `eslint` throws a parsing error:
```
Can not use keyword 'await' outside an async function
```

apepper: > We're using a top-level const only once, so inlining would be good enough.

It's true, that the const is only used once inside of `generateHtml`. But `generateHtml` is called multiple times. If we inline the code into `generateHtml` we would have to fetch the `asset-manifest.json` over and over again.

webpack.config.js
apepper: I did not use `inject: false` to get around the self closing tags issue. I did disable it, because `HtmlWebpackPlugin` would otherwise include _all_ webpack entry assets to the HTML file (e.g. `index.js`, `tracking.js` and `scrivito_extensions.js`).

With https://github.com/Scrivito/scrivito_example_app_js/pull/283/files#diff-149066e1413b1174eae4005fff6923c7R15 you can control much better where specific asset URLs are placed.

apepper:
REVIEW (5b39da6) (again)

src/prerenderContent/generateHtml.js
dcsaszar: I see. We could still avoid `then` by moving the get-JSON part into the function. But it will not make much of a difference, I guess.

webpack.config.js
dcsaszar: There's a misunderstanding here: My comment is not about `inject: false`, it is about the options object to the plugin in general.

1. In our HTML templates, we use self-closing link tags.
2. In the options, we tell the plugin to *not* use self-closing link tags.

This is an inconsistent setup.

dcsaszar: We checked that the setting does *not* change existing `link`s, so it's good enough.

PR #285, opened 2020-02-19
make Scrivito endpoint configurable via env like tenant or origin

toaster:


krishan:
Ideally, the CSP header should also be based on `SCRIVITO_ENDPOINT`, otherwise the user would have to remember to adjust this manually.

https://github.com/Scrivito/scrivito_example_app_js/blob/544d19ee6a3a3038a61e21d591ed021e9eb59260/public/_headersCsp.json#L7

apepper:
> Ideally, the CSP header should also be based on `SCRIVITO_ENDPOINT`, otherwise the user would have to remember to adjust this manually.

I created the PR https://github.com/Scrivito/scrivito_example_app_js/pull/287 to adjust CSP automatically.

PR #286, opened 2020-02-20
[webpack] Bugfix: Deep Clone headersCsp in devServerCspHeaders.

apepper:
Otherwise `devServerCspHeaders` modifies `headersCsp` directly and adds `"'unsafe-eval'"` to `script-src`!

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,6 +3,7 @@ const dotenv = require("dotenv");
 const path = require("path");
 const process = require("process");
 const webpack = require("webpack");
+const lodash = require("lodash");
 const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 const CopyWebpackPlugin = require("copy-webpack-plugin");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -204,7 +204,7 @@ function generatePlugins({ isProduction, isPrerendering, scrivitoOrigin }) {
 }
 
 function devServerCspHeader() {
-  const directives = Object.assign({}, headersCsp);
+  const directives = deepClone(headersCsp);
```
dcsaszar: * [ ] We don't need to build our own `lodash.cloneDeep(headersCsp);`

apepper: We currently only have `lodash-es` which AFAIK is not accessible in plain node - at least not in all node versions.

Since the method in question is a one-liner I think we can "roll our own".

dcsaszar: We're already using it in the generators, that's why I think it works.

apepper: The generators have a dedicated `package.json` file (see https://github.com/Scrivito/scrivito_example_app_js/blob/master/generator-scrivito/package.json#L17 ).

apepper:
REVIEW (99628f4)

PR #293, opened 2020-03-12
Js fingerprints

apepper:


src/_prerender_content.html
```diff
@@ -0,0 +1,6 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="<%= htmlWebpackPlugin.files.chunks.prerender_content.entry %>"></script>
```
apepper: * [ ] Still open: `storePrerenderedContent` should remove the fingerprinted `prerender_content.js` and `prerender_content.js.LICENSE.txt`.

webpack.config.js
```diff
@@ -108,7 +108,12 @@ function webpackConfig(env = {}) {
     },
     output: {
       publicPath: "/",
-      filename: "assets/[name].[contenthash].js",
+      filename: chunkData => {
```
dcsaszar: * [ ] Q: How about an `import '!!file-loader!...'` in the component? Moves the complexity to where it is used.

apepper:
REVIEW (a4e947a)

PR #296, opened 2020-03-25
Npm update

apepper:


apepper:
FYI: Upgrading `@babel/core` adds 17 kb more to the payload! Sticking to version `7.8.7` for now.

PR #297, opened 2020-03-25
Prettier 1 => 2 (+ eslint updated)

apepper:
See https://prettier.io/blog/2020/03/21/2.0.0.html for a detailed description of the new changes.

Biggest change:

> After a long and careful consideration, we decided to change the default values for the `trailingComma` (from `none` to `es5`), `arrowParens` (from `avoid` to `always`), and `endOfLine` (from `auto` to `lf`) options.

This codebase already used `trailingComma` `es5`.


apepper:
Strangly localy `npm run eslint` passes without any errors .

apepper:
REVIEW (989ad4a)

Prettier 2 does no longer support node 8. Upgraded to node 10 instead, since node 8 already reached End-of-life on 2019-12-31.

apepper:
REVIEW (6504b4a)

PR #298, opened 2020-03-25
html-webpack-plugin 3 => 4

apepper:
> **BREAKING CHANGES**
> The assets object which is used for the template parameters and inside hooks was changed. The chunks property was removed and the js and css property was converted from a string into an object `{ entryName: string, path: string}`

Source: https://github.com/jantimon/html-webpack-plugin/blob/master/CHANGELOG.md#400-2020-03-23


apepper:
REVIEW (f4fa810)

apepper:
> The comment is a bit lenghty and has a typo

Would this be better?
```js
inject: false, // needs custom order of imports
```

dcsaszar:
> needs custom order of imports

If we replace "imports" by "script tags" we're good.

apepper:
REVIEW (734aa55)

PR #300, opened 2020-03-26
Scroll to fragment

apepper:
Makes use of the npm package [`scroll-to-fragment`](https://www.npmjs.com/package/scroll-to-fragment) to support scrolling to fragments/hashes.

scroll-to-fragment_smaller

This PR also generates ids for all HeadlineWidgets based on the content of the headline, e.g. https://deploy-preview-300--scrivito-example.netlify.com/#our-latest-blog-news or https://deploy-preview-300--scrivito-example.netlify.com/#different-homepage-variants.

A `#footer` id is also set. E.g. https://deploy-preview-300--scrivito-example.netlify.com/#footer.

Since the example app now supports scrolling to anchors we can remove `react-scroll`, which was quite heavy for the functionality it offered. With this the file size of `index.js` decreased from `979 KiB` down to `958 KiB` (`21 KiB` saved).

Fixes https://github.com/infopark/scrivito_js/issues/3907.


apepper:
REVIEW (5aa83ca)

src/Widgets/HeadlineWidget/HeadlineWidget.scss
```diff
@@ -0,0 +1,9 @@
+.headline-widget {
+  // Fix navigation overlapping anchor issue
+  // See .navbar "min-height" value
+  // Source: https://stackoverflow.com/a/38106970/881759
+  &-anchor {
```
dcsaszar, line 5: In other widget style files we use the double-dash separator, for example https://github.com/Scrivito/scrivito_example_app_js/blob/master/src/Widgets/PriceWidget/PriceWidget.scss#L5

* [ ] Please check

apepper:
REVIEW (f5ef6d9)

PR #313, opened 2020-05-25
[csp] Added "frame-ancestors" policy.

apepper:
This is the newer version of the deprecated "X-Frame-Options" header.

This PR also white-lists `https://*.scrivito.com` which will soon be a new place to edit scrivito content.

### DIFF
--- a/public/_headersCsp.json
+++ b/public/_headersCsp.json
@@ -13,5 +13,5 @@
   ],
   "object-src": "'none'",
   "block-all-mixed-content": true,
-  "frame-ancestors": ["'self'", "https://edit.scrivito.com"]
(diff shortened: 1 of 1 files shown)


apepper:
> Did you verify the CSP rule takes precedence over this header?

Strait from the specification (https://www.w3.org/TR/CSP2/#frame-ancestors-and-frame-options):

> The `frame-ancestors` directive _obsoletes_ the `X-Frame-Options` header. If a resource has both policies, the `frame-ancestors` policy SHOULD be enforced and the `X-Frame-Options` policy SHOULD be ignored.

apepper:
REVIEW (6c7d166)

PR #318, opened 2020-06-09
[css] Corrected breakpoints for Navigation, Blog and ThumbnailGalleryWidget

### LINKED ISSUE #317: The menu layout looks strange at exactly 992px width
* Open the homepage, for example https://scrivito-example.netlify.app/
* In Chrome dev tools, select responsive mode and change the width to 992
### Expected

One of the layouts for 991 or 993 px.

### Actual

image

apepper:
Fixes #317.

Just for references:

`cat _responsive.scss`:

```scss
/* - BS3 variables - */

$screen-xs-max: 767px;
$screen-sm-min: 768px;
$screen-sm-max: 991px;
$screen-md-min: 992px;
$screen-md-max: 1199px;
$screen-lg-min: 1200px;
```

### DIFF
--- a/src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidget.scss
+++ b/src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidget.scss
@@ -99,7 +99,7 @@
   }
 }
 
-@media (max-width: $screen-sm-min) {
+@media (max-width: $screen-sm-max) {
   .thumbnail-gallery-widget {
     .thumbnail-gallery-widget--content {
       .fa {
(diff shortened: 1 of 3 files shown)


src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidget.scss
```diff
@@ -99,7 +99,7 @@
   }
 }
 
-@media (max-width: $screen-sm-min) {
+@media (max-width: $screen-sm-max) {
```
dcsaszar, line 102: This changes the responsive behaviour.
If this was an off-by-one, it should now be `screen-xs-max`.

* [ ] FYI It looks good to me, but please make sure that the change is intended.

apepper: This is actually an intended change by me. The design looks better to me, if we use all of `$screen-sm` and not just the lower bound.

PR #333, opened 2020-09-15
Update react-cookie-consent

pietruszka:


src/Components/CookieConsent.js
```diff
@@ -39,6 +39,8 @@ class CookieConsent extends React.Component {
         contentClasses="cookie-content d-flex align-items-center"
         buttonClasses="cookie-button btn btn-primary ml-auto mr-2"
         declineButtonClasses="cookie-button btn btn-secondary ml-auto"
+        buttonWrapperClasses="button-wrapper"
```
apepper, line 42:
```suggestion
        buttonWrapperClasses="button-wrapper ml-auto"
```

apepper (changes requested):
FEEDBACK (f83aef2)

Almost there. For mobile it looks already quite nice. Only on Desktop it does not look 100% optimal, since the buttons right next to the text. Maybe added some more CSS will fix the issue (see comment).

apepper (approved):
LGTM (f83aef2)

Good job 🚀 !

PR #335, opened 2020-09-16
Update fuse.js

pietruszka:


apepper (changes requested):
FEEDBACK (24c501a)

Very nice work 🚀 !

Now to make `es-check` happy we need to exclude `scrivito_extensions.<hex>.js` from the check. `scrivito_extensions` are allowed to include newer javascript, since it does not need to be IE11 compatible.

Simply do this in `package.json`:

```json
    "es-check": "npm run build && es-check es5 --not=build/assets/scrivito_extensions. './build/**/*.js'"
```

apepper (approved):
LGTM (2eb8bab)

Good job 🚀 !

PR #337, opened 2020-09-16
react-reveal => react-awesome-reveal

apepper:
`react-reveal` is no longer maintained since two years. `react-awesome-reveal` seems to be a more modern version.

I still have to test, if it works at least ok with IE11.


apepper:
This does not support IE11 out of the box. Since the payload is also higher then `react-reveal-iframe` I'm going to close this PR.

PR #338, opened 2020-09-16
update history

pietruszka:


apepper (approved):
LGTM (07d0592)

Good job 🚀 ! Also works well in IE11.

apepper:
On a closer look I can see, that `scrivito` does not yet support `history` version `5`. Given that I would not like to use multiple different versions of `history` for now.

Thank you anyway for the PR!

PR #330, opened 2020-09-10
update copy-webpack-plugin

pietruszka:

PR #339, opened 2020-09-17
[npm] Bugfix: Downgraded "copy-webpack-plugin" to version 6.0.0 for now.

apepper:
Version 6.0.1 and newer (including 6.1.0) do not process the "transform" option well. `CSP-DIRECTIVES-PLACEHOLDER` inside of `_headers` is not replaced well.

Broken since https://github.com/Scrivito/scrivito_example_app_js/pull/330.

I will update the PR once I created an issue on https://github.com/webpack-contrib/copy-webpack-plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2929,20 +2929,19 @@
       "integrity": "sha1-Z29us8OZl8LuGsOpJP1hJHSPV40="
     },
     "copy-webpack-plugin": {
-      "version": "6.1.0",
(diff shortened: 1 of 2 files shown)


dcsaszar (approved):
🙈

apepper:
AFAICS the issue seems to be, that `../public` and `../public/_headers` where referenced in different `patterns`. See #340 for more details.

PR #340, opened 2020-09-17
Make transform also work for copy-webpack-plugin 6.1.0.

apepper:
Follow-up of https://github.com/Scrivito/scrivito_example_app_js/pull/339.

If multiple patterns reference similar or identical files (e.g. "../public" and "../public/_headers") it is not well defined, which pattern will be executed - the first or the last? ~To avoid this confusion now "../public" is copied, but only "public/_headers" is transformed.~ TO avoid this confusion copying `../public` does explicitly exclude `../public/_headers`.

copy-webpack-plugin 6.0.1 most likely broke because of https://github.com/webpack-contrib/copy-webpack-plugin/pull/484.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -166,14 +166,18 @@ function generatePlugins({ isProduction, isPrerendering, scrivitoOrigin }) {
     new Webpackbar(),
     new CopyWebpackPlugin({
       patterns: [
-        { from: "../public", globOptions: { ignore: ignorePublicFiles } },
         {
-          from: "../public/_headers",
-          transform: (content) => {
-            const csp = builder({ directives: headersCsp });
-            return content
-              .toString()
-              .replace(/CSP-DIRECTIVES-PLACEHOLDER/g, csp);
+          from: "../public",
+          globOptions: { ignore: ignorePublicFiles },
+          transform: (content, absoluteFrom) => {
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -166,14 +166,18 @@ function generatePlugins({ isProduction, isPrerendering, scrivitoOrigin }) {
...
-          from: "../public/_headers",
-          transform: (content) => {
-            const csp = builder({ directives: headersCsp });
-            return content
-              .toString()
-              .replace(/CSP-DIRECTIVES-PLACEHOLDER/g, csp);
+          from: "../public",
+          globOptions: { ignore: ignorePublicFiles },
+          transform: (content, absoluteFrom) => {
+            if (absoluteFrom.endsWith("/public/_headers")) {
```
dcsaszar, line 173: This looks pretty complex. The solution I expected was to add `_headers` to the `ignore(d)PublicFiles`. Am I missing something?

* [ ] Please try to simplify.

dcsaszar (commented):
See comment for question regarding simplification.

webpack.config.js
apepper: Good idea!

apepper:
REVIEW (a9d7027)

dcsaszar (approved):
😃

PR #343, opened 2020-10-01
node-sass => (dart-)sass

apepper:
I did check the resulting css output. Besides some reordering and some minor arithmetic precision difference the output is identical.

### Background

The `sass` package is a distribution of [Dart Sass](https://github.com/sass/dart-sass), compiled to pure JavaScript with no native code or external dependencies.

> ### Why Dart?
> Dart Sass has replaced Ruby Sass as the canonical implementation of the Sass language. We chose Dart because it presented a number of advantages:
>
> * It's fast. The Dart VM is highly optimized, and getting faster all the time (for the latest performance numbers, see perf.md). It's much faster than Ruby, and close to par with C++.
> * ...

Source: https://github.com/sass/dart-sass#why-dart

The `node-sass` package is a wrapper around [LibSass](https://sass-lang.com/libsass), the C++ implementation of Sass.

> Node Sass supports the same API as this package and is also faster (although it's usually a little slower than Dart Sass). However, it requires a native library which may be difficult to install, and it's generally slower to add features and fix bugs.

Source: https://www.npmjs.com/package/sass#see-also

`sass-loader` recommends to use `sass`.

In the past `node-sass` had been a source of unmet `npm audit` reports. Often times it took quite a long time until a fix was available.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1567,7 +1567,6 @@
       "version": "3.1.1",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-3.1.1.tgz",
       "integrity": "sha512-mM8522psRCqzV+6LhomX5wgp25YVibjh8Wj23I5RPkPppSVSjyKD2A2mBJmWGa+KN7f2D6LNh9jkBCeyLktzjg==",
-      "optional": true,
       "requires": {
         "normalize-path": "^3.0.0",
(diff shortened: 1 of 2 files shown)


dcsaszar (approved):
👍  Roughly 10% speed decrease for the build step is acceptable, taking the described issues of the fast version into account.

PR #344, opened 2020-10-15
Replace ReactCookieConsent

pietruszka:


src/utils/cookieConsentManagement.js
```diff
@@ -0,0 +1,58 @@
...
+export function getCookieConsent() {
+  return localStorage.getItem("CookieConsent") || "undecided";
+}
+
+export function acceptCookieConsent() {
+  localStorage.setItem("CookieConsent", "accepted");
+
+  resolveCookieConsent();
+}
+export function declineCookieConsent() {
```
apepper, line 56: Please add a new line:

```suggestion

export function declineCookieConsent() {
```

src/utils/cookieConsentManagement.js
```diff
@@ -0,0 +1,58 @@
+import * as Scrivito from "scrivito";
+
+let resolveCookie;
+let resolveCookieConsentCalledBefore = false;
+
+const cookiePromise = new Promise((resolve) => {
+  resolveCookie = resolve;
+});
+
+export function resolveCookieConsent() {
```
apepper, line 10: If I'm not mistaken, this method no longer needed to be exported.

```suggestion
function resolveCookieConsent() {
```

apepper (changes requested):
FEEDBACK (ae8f97b)

Very nice work 🚀 ! This actually saves 25kb minified not-gzipped from the `index.js` file.

Please see my tiny comments.

Also one more thing:
* [ ] Please do not render the cookie consent during prerender - the buttons do not work while js is still loading and the visitor might have already agreed previously. For that visible should only be set to `true` in `componentDidMount` (or more the react hooks equivalent).

apepper (approved):
LGTM (c8ca881)

Good job 🚀 ! Very nice work!

src/Components/CookieConsent.js
```diff
@@ -1,47 +1,48 @@
...
-  render() {
-    const root = Scrivito.Obj.root();
+  React.useEffect(() => {
+    setVisible(getCookieConsent() === "undecided");
 
-    if (!root) {
-      return null;
-    }
+    cookieConsentGiven().then(() => {
+      setVisible(false);
```
dcsaszar, line 20: * [ ] FYI we're rerendering `null` for no reason in most cases (all except `undecided`). I don't know if this has any measurable effect.

src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
```diff
@@ -82,7 +82,7 @@ class ThumbnailGalleryComponent extends React.Component {
           <ReactBnbGallery
             show={this.state.lightboxIsOpen}
             backgroundColor="rgba(22,22,22,.9)"
-            zIndex={1000000} // same value as react-cookie-consent
+            zIndex={1000000} // same value as CookieConsent
```
dcsaszar, line 85: * [ ] FYI if we want to make this comment more helpful, we could point to the CSS wording `.cookie-box` or `cookie box`.

src/utils/cookieConsentManagement.js
```diff
@@ -0,0 +1,59 @@
+import * as Scrivito from "scrivito";
+
+let resolveCookie;
+let resolveCookieConsentCalledBefore = false;
+
+const cookiePromise = new Promise((resolve) => {
+  resolveCookie = resolve;
+});
+
+function resolveCookieConsent() {
```
dcsaszar, line 10: * [ ] According to https://stackoverflow.com/a/29491617 this is not needed. `resolveCookie()` should resolve only once, regardless of the number of calls.

dcsaszar (changes requested):
Mostly LGTM. See comments for things that should be double checked.

For the mid-term I have the rough idea of replacing the DIY cookie state management by using `useCallback` and `useContext` hooks.

src/Components/CookieConsent.js
apepper: I don't understand the question. What do you suggest or what is your concern?

dcsaszar: If we would like minimise the number of re-renders, we should call `setVisible` only if the arg differs from the current `visible` value.

apepper (approved):
LGTM (5458836)

Good job 🚀 !

dcsaszar (changes requested):
I like that the cookie consent handling is now covered by two dedicated modules instead of "utils".

However, `cookieConsentUrl` is misplaced. It is only used in the banner component.

* [ ] So it should be a private helper instead of being exported from `CookieConsentProvider` (where it is not used at all).

Exporting both `CookieConsentContext.Provider` and `CookieConsentProvider` is error prone. 
Assuming the implementation is based on https://kentcdodds.com/blog/application-state-management-with-react, which implements `useCount` as the "getter" hook,

* [ ] I suggest *not* to export `CookieConsentContext`, and `export {CookieConsentProvider, useCookieConsent}` instead. 

### Naming

Some names are a little general and could be clearer.

* [ ] Suggestions:
  * `CookieConsent` (the component) -> `CookieConsentBanner`
  * `cookieConsent` (from the context) -> `cookieConsentValue` / `cookieConsentChoice`

### Localization

The banner text was localised in ReactCookieConsent, which, assuming the authors took care of proper text, is why it was OK-ish not to be able to edit it. This is no longer the case. Sorry for only discovering that in my 2nd review:

* [ ] The wording of the banner and buttons should be editable, English fallback text is OK

apepper:
- [ ] Regarding `cookieConsentUrl`: Currently there is only one caller, but in a future PR this will most likely also be used by the Google Maps Widget. But for now we can move the function into the `CookieConsent(Banner)`.

- [ ] Regarding exporting `CookieConsentContext` - this is a good point, that we don't want to allow a second `CookieConsentContext.Provider`. Therefore exporting `useCookieConsent` that wraps `React.useContext(CookieConsentContext)` is a good idea.

- [ ] I'm open for the naming suggestions.

- Regarding localization - please let us not open but this pandora's box. This story already had quite a lot of iterations. I think making the cookie banner scrivito editable is a different story that should not be tackled in this PR - if at all. If someone really want's the cookie banner to be in german (or any other language) he/she needs to adjust the code for it.

src/App.js
```diff
@@ -8,7 +8,7 @@ import Footer from "./Components/Footer";
 import Intercom from "./Components/Intercom";
 import Navigation from "./Components/Navigation";
 import NotFoundErrorPage from "./Components/NotFoundErrorPage";
-import CookieConsent from "./Components/CookieConsent";
+import CookieConsentBanner from "./Components/CookieConsent";
```
apepper, line 11: Please also rename the file accordingly:

```js
import CookieConsentBanner from "./Components/CookieConsentBanner";
```

src/Components/CookieConsentContext.js
```diff
@@ -1,48 +1,39 @@
...
-  if (!root) {
-    return null;
-  }
-
-  const cookieConsentLink = root.get("cookieConsentLink");
-
-  if (!cookieConsentLink) {
-    return null;
-  }
+export function useCookieConsent() {
```
apepper, line 28: Let's keep all the logic ( `cookieConsentChoice`, `acceptCookieConsent` and `declineCookieConsent`) inside of `CookieConsentProvider`:

```js
export function useCookieConsent() {
  return React.useContext(CookieConsentContext);
}

apepper (changes requested):
FEEDBACK (35e5093)

Please see my comments.

apepper (approved):
LGTM (a975e11)

Good job 🚀 !

apepper:
@dcsaszar If there are still concerns we can address them in a follow up PR.

PR #345, opened 2020-10-17
Add arguments to generators

cbou:

PR #346, opened 2020-10-19
"yo scrivito:obj [<name>]" and "yo scrivito:widget [<name>]"

apepper:
Inspired by https://github.com/Scrivito/scrivito_example_app_js/pull/345.

### DIFF
--- a/generator-scrivito/app/index.js
+++ b/generator-scrivito/app/index.js
@@ -6,19 +6,19 @@ module.exports = class extends Generator {
     this.log("Welcome to the Scrivito Obj/Widget generator");
   }
 
-  start() {
-    this.prompt({
+  async start() {
+    const answers = await this.prompt({
       type: "list",
       name: "type",
       message: "What do you want to generate?",
       choices: ["An Obj class", "A Widget class"],
-    }).then((answers) => {
-      if (answers.type === "An Obj class") {
-        this.composeWith("scrivito:obj");
-      }
-      if (answers.type === "A Widget class") {
-        this.composeWith("scrivito:widget");
-      }
     });
+
+    if (answers.type === "An Obj class") {
+      this.composeWith("scrivito:obj");
+    }
+    if (answers.type === "A Widget class") {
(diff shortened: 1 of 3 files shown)


generator-scrivito/app/index.js
```diff
@@ -6,6 +6,12 @@ module.exports = class extends Generator {
     this.log("Welcome to the Scrivito Obj/Widget generator");
   }
 
+  usage() {
+    return `yo scrivito
+  yo scrivito:obj [<name>]
```
dcsaszar, line 11: Where does the `[<name>]` notation come from? For me it was misleading because I tried the literal.
```zsh
yo scrivito:obj [FaqPage]
```

* [ ] FYI Maybe rewrite this for the common case `yo scrivito:obj <name>` or use another standard form (man style): `yo scrivito:obj [name ...]`.

However, if the `[<>]` notation is widely accepted in the yeoman world, fine with me.

dcsaszar (approved):
Nice

generator-scrivito/app/index.js
apepper: That is the default notation of `yo`. E.g. `yo scrivito:obj --help` will show:

```Usage:
  yo scrivito:obj [<name>] [options]

Options:
  -h,   --help           # Print the generator's options and usage
        --skip-cache     # Do not remember prompt answers             Default: false
        --skip-install   # Do not automatically install dependencies  Default: false
        --force-install  # Fail on install dependencies error         Default: false

Arguments:
  name  # The name of the new Obj class (e.g. FaqPage)  Type: String  Required: false
```

dcsaszar: I see. If the options is formatted as `[options]` the names should be formatted as `[names]`. We could also be a bit clearer with sth like `[obj_class_names]`. Not sure about the casing, but it's only an optional idea.

apepper:
Thanks to @cbou for the original PR!

cbou:
Much better than my quick PR ;) Thanks @apepper !

PR #349, opened 2020-10-26
Extract StaticMap, transform GoogleMapsWidgetComponent into functional component

pietruszka:


src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
```diff
@@ -56,45 +56,45 @@ function StaticMap({ address, apiKey, zoom, children }) {
...
-        setMapBoundary({
-          elementHeight,
-          elementWidth,
-          height,
-          width,
-        });
-      }
-    }
-  }
+  }, [mapBoundary]);
```
dcsaszar, line 97: If we are rerunning the effect because the size changes, we also have to take care of removing the previous event listener.

* [ ] Q: Why is this change needed? The commit message doesn't give an explanation.

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
```diff
@@ -7,129 +7,135 @@ import "./GoogleMapsWidget.scss";
...
-    const elementHeight = this.outerDivRef.current.offsetHeight;
-
-    if (
-      this.state.elementWidth !== elementWidth ||
-      this.state.elementHeight !== elementHeight
-    ) {
-      let width = elementWidth;
-      let height = elementHeight;
+  return (
+    <InteractiveMap address={address} zoom={zoom} apiKey={apiKey}>
```
dcsaszar, line 26: * [ ] FYI because we have two identical code paths that only differ in the outer component we could use a variable component, for example:

```jsx
const Map = (mapType === "static") ? StaticMap : InteractiveMap;
<Map ... />
```

dcsaszar (changes requested):
Nice one. I found some room for improvement but also potential bugs. See comments.

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
pietruszka: `[mapBoundary]` is used to allow `handleResize` operates on a fresh state. With `[]`,  `handleResize` operates on a stale state with initial values. And we want to reduce the amount of invocation of `setMapBoundary`, with this condition ```
        if (
          mapBoundary.elementWidth !== elementWidth ||
          mapBoundary.elementHeight !== elementHeight
        ) ```

The previous event listener is removed in useEffect's return function. I've checked if there is any leak and I didn't noticed any problems:
```
> window.getEventListeners(window)
{message: Array(3), focus: Array(1), resize: Array(1), beforeunload: Array(1)}
```

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
```diff
@@ -7,129 +7,128 @@ import "./GoogleMapsWidget.scss";
...
-        height = Math.round(maxWidth * factor);
+function StaticGoogleMap({ address, apiKey, zoom, children }) {
+  const [mapSize, setMapSize] = React.useState({
+    elementHeight: 0,
+    elementWidth: 0,
+    height: null,
+    width: null,
+  });
+
+  const ref = React.useRef(null);
```
dcsaszar, line 49: Using `useRef` with `useEffect` is discouraged, because it may cause issues when passing the ref to child components.  ([Hooks FAQ](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node)).

* [ ] FYI I'd recommend to stick with the recommended `useCallback` approach to allow for code changes without running into subtle issues.

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
```diff
@@ -7,129 +7,128 @@ import "./GoogleMapsWidget.scss";
...
+        loading="lazy"
+      />
+      {children}
+    </div>
+  );
+}
 
-        const factor = elementHeight / elementWidth;
-        height = Math.round(maxWidth * factor);
+function StaticGoogleMap({ address, apiKey, zoom, children }) {
```
dcsaszar, line 41: FYI I'm not full happy with this roll-our-own size observer implementation.
* It has obvious shortcomings, for example no handling of reflow b/c of lazy CSS or content changes 
* It is rather complex, which may hide potential bugs

Given we already have a dependency to `resize-observer-polyfill` via `react-slick`, we could use a 3rd-party lib like `use-resize-observer` instead of roll-our-own. This would simplify our code by hopefully keeping the bundle size small.

* [ ] WDYT? We could make this a follow-up also.

dcsaszar (approved):
OK to merge. But please see my comments for potential improvements.

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
pietruszka: Sure, in a follow-up I'll implement suggestions

PR #352, opened 2020-11-03
Bugfix: Trim input on SearchResults page.

### LINKED ISSUE #351: Search with only a space breaks the page
http://localhost:8080/search-results?q=+

I got issues like:

```
Uncaught ClientError: Missing non-blank value (query term: {"field"=>"*", "operator"=>"contains_prefix", "value"=>" "})
    at errorForResponse (webpack:///../node_modules/scrivito/index.js?:9826:12)
    at handleAjaxResponse (webpack:///../node_modules/scrivito/index.js?:9789:17)
```

```
The above error occurred in the <SearchResultsComponent> component:
    in SearchResultsComponent (created by Scrivito.CurrentPage)
    in Scrivito.CurrentPage (created by App)
    in div (created by App)
    in div (created by App)
    in HelmetProvider (created by App)
    in CookieConsentProvider (created by App)
    in ErrorBoundary (created by App)
    in App

React will try to recreate this component tree from scratch using the error boundary you provided, ErrorBoundary.
```

apepper:
Fixes #351.

### DIFF
--- a/src/Objs/SearchResults/SearchResultsComponent.js
+++ b/src/Objs/SearchResults/SearchResultsComponent.js
@@ -15,7 +15,7 @@ class SearchResultsComponent extends React.Component {
   }
 
   calculateResults() {
-    if (!this.props.params.q) {
+    if (!this.query()) {
       return {
         searchResults: [],
         totalCount: 0,
@@ -23,11 +23,11 @@ class SearchResultsComponent extends React.Component {
       };
     }
 
-    let search = Scrivito.Obj.where(
-      "*",
-      "containsPrefix",
-      this.props.params.q
-    ).andNot("_objClass", "equals", blacklistObjClasses);
+    let search = Scrivito.Obj.where("*", "containsPrefix", this.query()).andNot(
+      "_objClass",
+      "equals",
+      blacklistObjClasses
+    );
 
     // make sure, that tags are calculated _before_ limiting to specific tag.
     const tags = search.facet("tags").map((tag) => tag.name());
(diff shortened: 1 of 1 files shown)


dcsaszar (approved):
LGTM

FYI Please note that `.trim()` is not available on IE~~11~~9 and triggers a new polyfill.
With an alternative form of for example `q.replace(/^\s*|\s*$/g, "")` the bundle becomes ~.6 KB smaller. On the other hand this is only .24 % savings overall.

apepper:
> FYI Please note that .trim() is not available on IE11 and triggers a new polyfill.

Strange: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim claims, that `.trim()` is supported since IE9.

dcsaszar:
Hmm, I have to admit I was only guessing after reading about issues with `trim` and IE on stackoverflow. Not sure what else could cause the change in package size. Maybe it's just random realignment caused by different compiler output.

apepper:
@dcsaszar You're right that the polyfill is taken. If I set `debug: true` for `@babel/preset-env` I do the the following output:

```
es.string.trim { "ie":"11" }
```

But I haven't figured out who is wrong: MDN or `preset-env`/`core-js`?

apepper:
`.trim()` does definitely not work on IE9:

trim on IE9

dcsaszar:
Works on IE11 though 🤷 
image
([test source](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/String/prototype/trim/detect.js))

apepper:
It also works on IE10 😄 :

trim on IE10

So MDN is wrong with the claim, that trim already works with IE9 and core-js is wrong that it claims, that IE does not support trim at all.

I created two PRs ([1](https://github.com/mdn/browser-compat-data/pull/7254)  and [2](https://github.com/zloirock/core-js/pull/881)) to fix both sources.

apepper:
@dcsaszar I suggest to merge this PR as is and hope that the package size will decrease again, once `core-js-compat` levels up.

PR #353, opened 2020-11-03
Replace resize with hook

pietruszka:
Follow-up to https://github.com/Scrivito/scrivito_example_app_js/pull/349#discussion_r515858241

https://www.npmjs.com/package/use-resize-observer#the-onresize-callback
> By the default the hook will trigger a re-render on all changes to the target element's width and / or height.

That's why we don't have to compare anymore an old and a new size.

### DIFF
--- a/src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
+++ b/src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
@@ -1,5 +1,6 @@
 import * as React from "react";
 import * as Scrivito from "scrivito";
+import useResizeObserver from "use-resize-observer";
 
 import googleMapsApiKey from "../../utils/googleMapsApiKey";
 import googleMapsImageUrl from "../../utils/googleMapsImageUrl";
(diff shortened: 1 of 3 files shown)


pietruszka:
@dcsaszar Please review

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
```diff
@@ -109,6 +63,20 @@ function StaticGoogleMap({ address, apiKey, zoom, children }) {
...
+  let newHeight = height;
+
+  if (newWidth > maxWidth) {
+    newWidth = maxWidth;
+
+    const factor = height / width;
+    newHeight = Math.round(maxWidth * factor);
+  }
+
+  return { width: newWidth, height: newHeight };
```
apepper, line 77: Now that this is a dedicated method we can simplify it further:

```suggestion
  if (width > maxWidth) {
    const factor = height / width;
    const adjustedHeight = Math.round(maxWidth * factor);

    return { width: maxWidth, height: adjustedHeight };
  }

  return { width, height };
```

src/Widgets/GoogleMapsWidget/GoogleMapsWidgetComponent.js
```diff
@@ -109,6 +63,17 @@ function StaticGoogleMap({ address, apiKey, zoom, children }) {
   );
 }
 
+function scaleMapSize(width, height) {
+  if (width > maxWidth) {
```
dcsaszar, line 67: * [ ] FYI the input can be `null` according to line 43. This function works by change with `null` input, but we may want to make it more explicit. Also, with `0` input the function may output `NaN` values. I think with a guard like `!height || !width` we would avoid any potential headache.

dcsaszar (approved):
LGTM. See comment.

dcsaszar (approved):

apepper (commented):
FEEDBACK (a696188)

Nice work 🚀 ! Almost there!

I found two places that could be improved.

apepper (approved):
LGTM (897bb51)

Good job 🚀 !

PR #356, opened 2020-11-19
Bugfix: No longer crash SearchResultsComponent, if params.q is undefined

apepper:
The error can be observed, if one calls "npm run prerender".

Introduced with https://github.com/Scrivito/scrivito_example_app_js/pull/352.

### DIFF
--- a/src/Objs/SearchResults/SearchResultsComponent.js
+++ b/src/Objs/SearchResults/SearchResultsComponent.js
@@ -87,7 +87,7 @@ class SearchResultsComponent extends React.Component {
   }
 
   query() {
-    return this.props.params.q.trim();
+    return this.props.params.q ? this.props.params.q.trim() : "";
   }
 }
 


src/Objs/SearchResults/SearchResultsComponent.js
```diff
@@ -87,7 +87,7 @@ class SearchResultsComponent extends React.Component {
   }
 
   query() {
-    return this.props.params.q.trim();
+    return this.props.params.q ? this.props.params.q.trim() : "";
```
dcsaszar, line 90: * [ ] FYI/Q: Can we shorten this to `this.props.params.q?.trim();`?

dcsaszar (approved):
LGTM

src/Objs/SearchResults/SearchResultsComponent.js
apepper: If we simple do `this.props.params.q?.trim();` `query()` could return `undefined`. Components of inner props actually expect string values, e.g. `SearchResultItem`: https://github.com/Scrivito/scrivito_example_app_js/blob/a4e671110e0958e3071409f795775f3bdd2e6340/src/Objs/SearchResults/SearchResultItem.js#L50-L51

dcsaszar: FYI If there is no query this path is not taken.

PR #357, opened 2020-11-19
Migrate to edit.scrivito.com

pietruszka:


apepper (changes requested):
FEEDBACK (34bfb1a)

Good stuff 🚀 !

I only found one more place, that should be improved: `public/robots.txt` still references `/scrivito/`. I think we can simply remove the line an keep the - more or less empty - `robots.txt` around.

apepper (approved):
LGTM (34bfb1a)

After some more thoughts we should leave `/scrivito/` inside of `robots.txt`. Those routes still work but should never be indexed by search engines.

PR #359, opened 2020-12-09
Update eslint@7.15.0 and other related packages

pietruszka:
Added new rules:
- exporting default anonymous function is not allowed (https://github.com/benmosher/eslint-plugin-import/blob/v2.22.1/docs/rules/no-anonymous-default-export.md)
- naming non-hook function with `use*` prefix is not allowed (https://github.com/facebook/react/issues/17636#issuecomment-567081987)
- if a function contains only a return statement, braces are not allowed (https://eslint.org/docs/rules/arrow-body-style)


apepper (approved):
LGTM (e8785cf)

PR #363, opened 2021-01-08
Force script to exit if chrome can't start

cbou:
I try to use AWSAmplify and the prerender script never ends without giving any hint on what is going on.
This was very confusing...


apepper:
Thank you so much for your contribution! It is really nice to see, when someone improves the example app.

I took you ideas, refactored the code a bit and submitted it as #364. @cbou Feel free to leave a comment or review on #364.

PR #364, opened 2021-01-13
Errors during prerender stop script

apepper:
Original patch by https://github.com/cbou (closes #363).

### DIFF
--- a/prerender/storePrerenderedContent.js
+++ b/prerender/storePrerenderedContent.js
@@ -46,38 +46,42 @@ async function storePrerenderedContent() {
   const server = await startServer(SOURCE_DIR);
   log("🗄️  Express server started...");
 
-  log("🖥️️  Starting browser...");
-  const browser = await puppeteer.launch();
-  log("🖥️️  Browser started");
+  try {
+    log("🖥️️  Starting browser...");
+    const browser = await puppeteer.launch();
+    log("🖥️️  Browser started");
 
-  const url = "http://localhost:8080/_prerender_content.html";
-  log(`🖥️️  Visiting ${url} ...`);
-  const page = await visitUrl(browser, url);
+    try {
+      const url = "http://localhost:8080/_prerender_content.html";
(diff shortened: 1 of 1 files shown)


prerender/storePrerenderedContent.js
```diff
@@ -86,6 +86,15 @@ function log(message, ...args) {
   console.log(`[storePrerenderedContent] ${message}`, ...args);
 }
 
+async function startBrowser() {
+  try {
+    return await puppeteer.launch();
+  } catch (e) {
+    reportError("Could not start browser!", e);
+    process.exit(1);
```
dcsaszar, line 94: IIRC, using `process.exit` may lead to loss of log output. So, this fix kinda defeats the purpose. We should use the same strategy as below, with `process.exitCode = 1;` and ending the script gracefully.

dcsaszar (changes requested):
See comment. I hate this subtle `process.exit` issue, but let's try to avoid the pitfall.

prerender/storePrerenderedContent.js
apepper: Good point.

apepper:
REVIEW (11f9053)

I did a closer analysis of the error. Node would not exit the process with the original `process.exitCode = 1` implementation, since an express server would still run.

I forced pushed a new version, that will always close the browser and the server. This is regardless of errors in between. With that node should exit as expected.

dcsaszar (approved):
LGTM.

* [ ] Q: How does this change fix the original complaint "no useful log if the browser failed"?

apepper:
> Q: How does this change fix the original complaint "no useful log if the browser failed"?

My guess is, that `AWSAmplify` only provided logs after a process was fully done. Since the process never terminated it did not output that message.

But I did not test this myself. I only tested it locally with a invalid `executablePath`.

cbou:
Thanks for your PR! I think it's should be fine for AWS.

PR #365, opened 2021-01-13
Added GroupWidget

apepper:
It is a simple div container that can hold multiple widgets.

GroupWidget

### DIFF
--- a/src/Widgets/GroupWidget/GroupWidgetEditingConfig.js
+++ b/src/Widgets/GroupWidget/GroupWidgetEditingConfig.js
@@ -1,0 +1,7 @@
+import * as Scrivito from "scrivito";
+import groupWidgetIcon from "../../assets/images/group_widget.svg";
+
+Scrivito.provideEditingConfig("GroupWidget", {
+  title: "Group",
+  thumbnail: groupWidgetIcon,
+});
--- a/src/assets/images/group_widget.svg
+++ b/src/assets/images/group_widget.svg
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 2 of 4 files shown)


src/assets/images/group_widget.svg
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
```
dcsaszar, line 9: * [ ] FYI if saving as SVG instead of Inkscape SVG there will be less metadata. Alternatively we could use SVGOMG.

src/Widgets/GroupWidget/GroupWidgetEditingConfig.js
```diff
@@ -0,0 +1,7 @@
+import * as Scrivito from "scrivito";
+import groupWidgetIcon from "../../assets/images/group_widget.svg";
+
+Scrivito.provideEditingConfig("GroupWidget", {
+  title: "Group",
```
dcsaszar, line 5: * [ ] FYI I see we aren't using `description` yet. I wonder if this is a good candidate. As the widget in itself doesn't add anything to the content, we could explain it using `description`.
On the other hand, the widget's simplicity makes it kinda self-explanatory, so feel free to go without description.

dcsaszar (approved):
Nice and simple. See comments.

src/assets/images/group_widget.svg
apepper: Non of the (editor-only) svg's have yet been optimized. I see this as a separate task.

src/Widgets/GroupWidget/GroupWidgetEditingConfig.js
apepper: I think it's good enough.

src/assets/images/group_widget.svg
dcsaszar: FYI I had checked the existing SVGs for `xmlns:inkscape`. There were none.

PR #366, opened 2021-01-15
Scrivito 1.19.0 + add backend priority flag

pietruszka:


src/config/scrivito.js
```diff
@@ -10,4 +10,8 @@ if (process.env.SCRIVITO_ENDPOINT) {
   config.endpoint = process.env.SCRIVITO_ENDPOINT;
 }
 
+if (process.env.SCRIVITO_PRERENDER === "true") {
```
apepper, line 13: In other places `process.env.SCRIVITO_PRERENDER` is plainly used. See for example https://github.com/Scrivito/scrivito_example_app_js/blob/2e59bcef9a80a1278596eea48dd748174e496dda/webpack.config.js#L30

Given that I would prefer a stripped down version:

```suggestion
if (process.env.SCRIVITO_PRERENDER) {
```

apepper (commented):
FEEDBACK (8e25f3a)

Nice 🚀 ! I only have one small complaint.

apepper (approved):
LGTM (e40951e)

Good job 🚀 !

apepper:
:exclamation: This `priority: "background"` implementation in the example app contained a bug. If you prerender with this bug every request will be of `priority: "background"`, not just during prerender. :exclamation:

This bug is fixed with Scrivito/scrivito_example_app_js#379.

PR #369, opened 2021-01-20
Allow to pass arguments to puppeteer

cbou:
During our migration to AWS Amplify we had the issue that AWS builds with the root user. The is not possible without the flag `--no-sandbox` (I know it bad :/ ). So I give the ability to pass arguments to Puppeteer. That could be useful for other issues.

### DIFF
--- a/prerender/storePrerenderedContent.js
+++ b/prerender/storePrerenderedContent.js
@@ -48,7 +48,9 @@ async function storePrerenderedContent() {
 
   try {
     log("🖥️️  Starting browser...");
-    const browser = await puppeteer.launch();
+    const browser = await puppeteer.launch({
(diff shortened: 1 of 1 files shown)


apepper:
First of all, thank you for your contribution.

I'm hesitant to merge this, because args for chromium can be simple flags (e.g. `--no-sandbox`) but also valued flags (see complete list under https://peter.sh/experiments/chromium-command-line-switches/). Given the `.split(" ")` this patch would only work for simple flags, but not valued flags.

To make a better suggestion I did provide another PR: #373. It enabled the needed flags, when running on AWS Lambda. @cbou Could you confirm, if #373 also fixes your troubles on AWS Amplify?

apepper:
Closing since with #387 we're no longer using the browser environment to prerender.

PR #373, opened 2021-01-29
[prerender] Disable sandbox in AWS lambda environment.

apepper:
Otherwise prerendering might not work.

See [1] for details about env "AWS_EXECUTION_ENV".

[1] https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

Motivated by https://github.com/Scrivito/scrivito_example_app_js/pull/369.

@cbou: Can you confirm, that prerendering works on "AWS Amplify" with this patch?

### DIFF
--- a/prerender/storePrerenderedContent.js
+++ b/prerender/storePrerenderedContent.js
@@ -48,7 +48,11 @@ async function storePrerenderedContent() {
 
   try {
     log("🖥️️  Starting browser...");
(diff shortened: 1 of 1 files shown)


dcsaszar (approved):
LGTM

apepper:
Closing since with #387 we're no longer using the browser environment to prerender.

PR #377, opened 2021-03-15
Use configureObjClassForContentType

apepper:
IF you upload a video to the Content Browser it now automatically becomes a "Video" obj class. Previously it was a generic "Download" obj class. See https://www.scrivito.com/js-sdk/configureObjClassForContentType for more details.

Because of that, we can restrict `VideoWidget` and `navigationBackgroundImage` more to `Video` or `Image`/`Video`.

### DIFF
--- a/src/config/scrivito.js
+++ b/src/config/scrivito.js
@@ -15,9 +15,3 @@ if (process.env.SCRIVITO_PRERENDER) {
 }
 
 Scrivito.configure(config);
-
-Scrivito.configureObjClassForContentType({
-  "image/*": "Image", // Factory default
-  "video/*": "Video",
-  "*/*": "Download", // Factory default
-});
--- a/src/config/scrivito.js
+++ b/src/config/scrivito.js
@@ -15,3 +15,9 @@ if (process.env.SCRIVITO_PRERENDER) {
 }
 
 Scrivito.configure(config);
+
+Scrivito.configureObjClassForContentType({
+  "image/*": "Image", // Factory default
+  "video/*": "Video",
+  "*/*": "Download", // Factory default
+});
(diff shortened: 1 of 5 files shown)


src/config/scrivito.js
```diff
@@ -15,3 +15,9 @@ if (process.env.SCRIVITO_PRERENDER) {
 }
 
 Scrivito.configure(config);
+
+Scrivito.configureObjClassForContentType({
+  "image/*": "Image", // Factory default
```
dcsaszar, line 20: * [ ] FYI not sure if the "Factory default" comments add value here, or are more confusing than helpful. For a user of the Example App the factory default is this file. How about dropping the comments?

src/config/scrivito.js
```diff
@@ -15,3 +15,9 @@ if (process.env.SCRIVITO_PRERENDER) {
 }
 
 Scrivito.configure(config);
+
+Scrivito.configureObjClassForContentType({
```
dcsaszar, line 19: * [ ] FYI the config file structure is less clear now.

**Before:**
* `Scrivito.configure` lives in `config/scrivito`.
* `Scrivito.configureContentBrowser` lives in `config/scrivitoContentBrowser`.

**Now:**
* `Scrivito.configure` lives in `config/scrivito`,
* `Scrivito.configureContentBrowser` lives in `config/scrivitoContentBrowser`.
* `Scrivito.configureObjClassForContentType` lives in `config/scrivito`.
 
As the reader, I wonder what is the rule here?

dcsaszar (approved):
Nice. Two comments for your consideration.

apepper:
REVIEW (ded85fe)

dcsaszar (approved):
❤️

PR #379, opened 2021-03-22
Bugfix: Only use background priority during prerender.

apepper:
The `priority` allows to differentiate between `foreground` and `background`. If set to `background` a different rate limit is used than with `foreground`. This should ensure, that prerendering tasks do not "eat" rate of visitors. See https://www.scrivito.com/js-sdk/configure for more details.

For this to work visitors must _not_ use `background` for the `priority`. Since https://github.com/Scrivito/scrivito_example_app_js/pull/366 that was sadly the case, if you prerendered the site.

This PR fixes the issue by only setting `background` during actual prerender (`src/prerender_content.js`) and _not_ during "regular" visits (`src/index.js`).

Diff best reviewed without whitespace changes.

### DIFF
--- a/src/config/history.js
+++ b/src/config/history.js
@@ -2,8 +2,10 @@ import { createBrowserHistory } from "history";
 import { useHistory as scrivitoUseHistory } from "scrivito";
 import { scrollToFragment } from "scroll-to-fragment";
 
-const history = createBrowserHistory();
+export function configureHistory() {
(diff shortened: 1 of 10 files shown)


dcsaszar (approved):

apepper:
REVIEW (7543697)

dcsaszar (approved):
👍

PR #380, opened 2021-03-29
Make configureScrivito side-effect free

dcsaszar:
This is general best practice, but also handy when running the code under Node.js in conjunction with `dotenv`.

### DIFF
--- a/src/config/scrivito.js
+++ b/src/config/scrivito.js
@@ -1,16 +1,16 @@
 import * as Scrivito from "scrivito";
 
-const config = { tenant: process.env.SCRIVITO_TENANT, adoptUi: true };
-
-if (process.env.SCRIVITO_ORIGIN) {
-  config.origin = process.env.SCRIVITO_ORIGIN;
-}
-
-if (process.env.SCRIVITO_ENDPOINT) {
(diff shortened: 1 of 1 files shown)


src/config/scrivito.js
```diff
@@ -1,16 +1,16 @@
...
-if (process.env.SCRIVITO_ENDPOINT) {
-  config.endpoint = process.env.SCRIVITO_ENDPOINT;
-}
+  if (process.env.SCRIVITO_ENDPOINT) {
+    config.endpoint = process.env.SCRIVITO_ENDPOINT;
+  }
 
-export function configureScrivito(options) {
   const priority = options?.priority;
   Scrivito.configure(priority ? { ...config, priority } : config);
```
apepper, line 15: Since we're now building a new "config" on each call, we can modify it directly.

Something like:

```js
if (options?.priority) {
  config.priority = options?.priority;
}
```

apepper (commented):
FEEDBACK (2a96dc6)

Nice 🚀 ! Please see my comment to further improve the code-flow.

apepper (approved):
LGTM (2cb9b76)

Nice 🚀 !

PR #381, opened 2021-03-29
Better SSR support

dcsaszar:
Three smaller changes to be able to render server side, even without a `window`.

**Note**
I also considered moving `configureScrivitoContentBrowser()` and `configureObjClassForContentType()` to the window specific block. These are not strictly needed in a window-less (and thus no-editing) environment.

### DIFF
--- a/src/Components/CookieConsentContext.js
+++ b/src/Components/CookieConsentContext.js
@@ -3,9 +3,11 @@ import * as React from "react";
 const CookieConsentContext = React.createContext({});
 
 export function CookieConsentProvider(props) {
-  const [cookieConsentChoice, setCookieConsentChoice] = React.useState(() =>
-    getCookieConsent()
+  const [cookieConsentChoice, setCookieConsentChoice] = React.useState(
+    "undecided"
   );
+
+  React.useEffect(() => setCookieConsentChoice(getCookieConsent()), []);
 
   function updateCookieConsent(consent) {
     localStorage.setItem("CookieConsent", consent);
--- a/src/utils/fullScreenWidthPixels.js
+++ b/src/utils/fullScreenWidthPixels.js
@@ -1,5 +1,9 @@
 import devicePixelRatio from "./devicePixelRatio";
 
+/**
+ * Based on (and thus optimized for) the Lighthouse mobile emulation.
+ * See https://github.com/GoogleChrome/lighthouse/blob/777bf1147fd0f6aca16ffefde1350bf6297476d4/lighthouse-core/config/constants.js#L58-L63.
+ * */
 const SSR_FALLBACK_WIDTH_PX = 945;
 
 function fullScreenWidthPixels() {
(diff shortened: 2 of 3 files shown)


src/Components/CookieConsentContext.js
```diff
@@ -3,10 +3,12 @@ import * as React from "react";
...
 
 export function CookieConsentProvider(props) {
-  const [cookieConsentChoice, setCookieConsentChoice] = React.useState(() =>
-    getCookieConsent()
+  const [cookieConsentChoice, setCookieConsentChoice] = React.useState(
+    "undecided"
   );
 
+  React.useEffect(() => setCookieConsentChoice(getCookieConsent()), []);
+
```
apepper, line 11: How about to extend `useCookieConsent` that it returns "undecided" `if (typeof window !== "undefined")`? With that `cookieConsentChoice` would be correct from the first render and not only after `React.useEffect`.

src/utils/fullScreenWidthPixels.js
```diff
@@ -1,6 +1,12 @@
 import devicePixelRatio from "./devicePixelRatio";
 
+const SSR_FALLBACK_WIDTH_PX = 945;
```
apepper, line 3: Could you add your commit comment in code? Not every reader will easily find the commit that introduced this number.

apepper (changes requested):
FEEDBACK (dd705bf)

Nice 🚀 !

> I also considered moving configureScrivitoContentBrowser() and configureObjClassForContentType() to the window specific block. These are not strictly needed in a window-less (and thus no-editing) environment.

I think not calling `configureScrivitoContentBrowser` in node is fine. But I would keep `configureObjClassForContentType` always active. As soon as we _some day_ add the option to create/upload binaries in node this option is useful.

src/Components/CookieConsentContext.js
dcsaszar: Given a hydrated DOM should be *exactly* the same SSR and CSR, it is advantageous *not* to initialize the state with "undecided". In other words: since SSR _always_ is `undecided`, the first render _always must be_ undecided, too.

src/utils/fullScreenWidthPixels.js
```diff
@@ -1,5 +1,9 @@
 import devicePixelRatio from "./devicePixelRatio";
 
+/**
+ * This value used is based on (and thus optimized for) the Lighthouse mobile emulation.
```
apepper, line 4: FYI: "This value used is based on" seems off. Maybe "This value is based on"? Or maybe shorten it to `Based on (and thus optimized for) the Lighthouse mobile emulation.`

apepper (commented):
FEEDBACK (18e78fd)

Almost there 🚀 !

src/utils/fullScreenWidthPixels.js
dcsaszar: True. Will force push.

apepper (approved):
LGTM (83bff9d)

Good job 🚀 !

apepper:
Feel free to merge, once travis is green.

PR #383, opened 2021-04-20
Prerendering: Fix <LS> and <PS> line terminators for IE11

dcsaszar:
Fixes a bug where a dump containing one of the line terminators `<LS>` and `<PS>` (see https://262.ecma-international.org/5.1/#sec-7.3) would lead to invalid code on older browsers (👋 IE11).

As an additional optimization, we allow to garbage collect the dump string after use.

### DIFF
--- a/src/prerenderContent/generatePreloadDump.js
+++ b/src/prerenderContent/generatePreloadDump.js
@@ -1,3 +1,13 @@
 export default function generatePreloadDump(preloadDump) {
-  return `window.preloadDump = ${JSON.stringify(preloadDump)};`;
+  return `window.preloadDump = ${stringLiteral(preloadDump)};`;
+}
+
+/**
+ * JSON may contain line separators, which are invalid in string literals in older browsers.
+ * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#issue_with_plain_json.stringify_for_use_as_javascript
+ * */
+function stringLiteral(string) {
+  return JSON.stringify(string)
+    .replace(/\u2028/g, "\\u2028")
+    .replace(/\u2029/g, "\\u2029");
 }
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,7 @@ configure();
 
 if (window.preloadDump) {
   Scrivito.preload(window.preloadDump).then(({ dumpLoaded }) => {
+    delete window.preloadDump;
     dumpLoaded ? hydrateApp() : renderApp();
   });
 } else {


apepper:
Some more background information. If you content by chance contains a U+2028 LINE SEPARATOR or U+2029 PARAGRAPH SEPARATOR it might break prerendering for browsers such as IE11. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#issue_with_plain_json.stringify_for_use_as_javascript for more details.

src/prerenderContent/generatePreloadDump.js
```diff
@@ -1,3 +1,13 @@
...
+}
+
+/**
+ * JSON may contain line separators, which are invalid in string literals in older browsers.
+ * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#issue_with_plain_json.stringify_for_use_as_javascript
+ * */
+function stringLiteral(string) {
+  return JSON.stringify(string)
+    .replace(/\u2028/g, "\\u2028")
+    .replace(/\u2029/g, "\\u2029");
```
apepper, line 12: How does this behave with newer browsers, that already implement https://github.com/tc39/proposal-json-superset ? Will a later `JSON.parse` of such a string still produce the same result?

apepper (commented):
FEEDBACK (19daefd)

Nice work 🚀!

I only have one open question. Please see my comment.

src/prerenderContent/generatePreloadDump.js
dcsaszar: Yes, every \uHHHH (4 hex digits) is a valid backslash-UnicodeEscapeSequence according to https://262.ecma-international.org/5.1/#sec-7.8.4

dcsaszar: Also note, it is not about JSON.parse, but parsing source code.

apepper: I know, that this is about parsing source code. But that parsed source code will eventually be put into a `JSON.parse` (by `Scrivito.preload`). So I want to make sure, that our bugfix does not introduce a new bug for engines, that already implement https://github.com/tc39/proposal-json-superset.

dcsaszar:
```js
eval(`"${String.fromCodePoint(0x2028)}"`) === "\u2028"
// true
```

apepper (approved):
LGTM (19daefd)

Very nice 🚀 !

PR #386, opened 2021-04-23
New baseline: Node 12

apepper:
Node 10 reaches EOL this month. Node 12 will be supported for the next year.


.travis.yml
```diff
@@ -1,7 +1,5 @@
 ---
 language: node_js
-node_js: 10
```
apepper, line 3: Travis now picks up the node version from `.nvmrc`.

dcsaszar (changes requested):
* [ ] We should also update https://github.com/Scrivito/scrivito_example_app_js/blob/master/package.json#L6

apepper:
REVIEW (8b2f3ae)

PR #389, opened 2021-04-26
Fix #382

### LINKED ISSUE #382: Links in Blog Overview Widgets don't adjust color according to background-image
When using the `BlogOverviewWidget` with a background-image selected, the font of the links don't adjust their color.

Bildschirmfoto 2021-04-15 um 12 29 12
Bildschirmfoto 2021-04-15 um 12 40 16

When the Widget is wrapped inside a section, the text is not displayed correctly. It seems to adjust to the color of the background image, not the color of the section. Although I am not sure the widget is supposed to be used that way. 

Bildschirmfoto 2021-04-15 um 12 44 34

The color of the font seems to be controlled by the css-class `.bg-dark-image`

agessler:
Regarding `Links in Blog Overview Widgets don't adjust color according to background-image #382 `

I found two issues. Blog panels are now always with white bachground and grey text. 
Secound issue was when card boxes was on dark background.. fa icons inside was white.. new grey.
bg-issue1

bg-issue2

### DIFF
--- a/src/assets/stylesheets/_blog.scss
+++ b/src/assets/stylesheets/_blog.scss
@@ -52,6 +52,8 @@
         }
 
         &--panel {
+          background: #fff;
+
           &:before {
             border-left-width: 0;
             border-right-width: 15px;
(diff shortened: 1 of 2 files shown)


apepper (approved):
LGTM (f411df3)

Very nice 🚀 ! Thanks for the fix.

PR #391, opened 2021-04-28
Simplify webpack config by separating development and production

dcsaszar:
We now have:

* `webpack.config.js` for the development setup. It is agnostic of a production setup. To keep the EnvironmentPlugin use DRY, it still looks at `env.production` once, but this is an official webpack env. Alternative ideas welcome.
* `webpack.config.production.js` for production. It relies on the dev setup to provide sensible defaults. The production settings are then added to the default, and filtered where needed.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,8 +14,6 @@ const headersCsp = require("./public/_headersCsp.json");
 // load ".env"
 dotenv.config();
 
-const isPrerendering = process.env.SCRIVITO_PRERENDER;
-
 let scrivitoOrigin = "";
 if (process.env.CONTEXT === "production") {
   scrivitoOrigin = process.env.URL;
@@ -46,7 +44,11 @@ function webpackConfig(env = {}) {
   return {
     mode: "development",
     context: path.join(__dirname, "src"),
-    entry: generateEntry(),
+    entry: {
+      index: "./index.js",
+      tracking: "./tracking.js",
+      scrivito_extensions: "./scrivito_extensions.js",
+    },
     target: "web",
     module: {
       rules: [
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -4,18 +4,25 @@ const path = require("path");
...
 const { WebpackManifestPlugin } = require("webpack-manifest-plugin");
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const Webpackbar = require("webpackbar");
-const ZipPlugin = require("zip-webpack-plugin");
 const headersCsp = require("./public/_headersCsp.json");
 
 // load ".env"
 dotenv.config();
 
+const isPrerendering = process.env.SCRIVITO_PRERENDER;
```
apepper, line 17: I think we should move all the prerendering logic to `webpack.config.production.js`. The production script already reads out `SCRIVITO_PRERENDER`. This would also remove more switches in `webpack.config.js`, so that `generateEntry` and `generatePlugins` could be inlined.

webpack.config.production.js
```diff
@@ -0,0 +1,41 @@
...
+
+const devWebpackConfig = require("./webpack.config");
+
+const isPrerendering = process.env.SCRIVITO_PRERENDER;
+
+function webpackConfig(env = {}) {
+  const devConfig = devWebpackConfig({ ...env, production: true });
+
+  return {
+    ...devConfig,
```
apepper, line 14: I think we should also _not_ make use of `devServer` in production/prerendering.

apepper (changes requested):
FEEDBACK (497c146)

Nice 🚀 ! Looks already much cleaner.

Please see my comments for further suggestions/improvements. I think we should move all the prerendering logic to production to keep development nice and clean. And since very soon prerendering will no longer need the browser an interactive browser session with the current prerendering code is no longer needed.

webpack.config.production.js
dcsaszar: Yeah. We'll lose a simple way to test changes on IE11, but I guess nobody will miss that.

webpack.config.js
```diff
@@ -4,18 +4,23 @@ const path = require("path");
...
 // load ".env"
 dotenv.config();
 
+let scrivitoOrigin = "";
+if (process.env.CONTEXT === "production") {
+  scrivitoOrigin = process.env.URL;
+} else if (process.env.DEPLOY_PRIME_URL) {
+  scrivitoOrigin = process.env.DEPLOY_PRIME_URL;
+}
+
```
apepper, line 23: FYI: Given that `scrivitoOrigin` is now only used in `webpackConfig` we can move this block back to its original place.

apepper (approved):
LGTM (5d9da7b)

Very nice 🚀 ! Only one FYI.

apepper (approved):
LGTM (489c39a)

Great stuff!

PR #393, opened 2021-04-30
Run eslint on webpack configs

apepper:
With two small exceptions we can run eslint on webpack configs.


dcsaszar (approved):
* [ ] FYI If we were to put the exceptions here it would not clutter the source files and be more DRY:

`.eslintrc`
```js
"overrides": [
    {
      "files": ["webpack.config.*"],
      "rules": {
        "import/no-commonjs": 0,
        "import/no-extraneous-dependencies": 0
      }
    },
```

apepper:
REVIEW (65260d2)

PR #394, opened 2021-05-06
Added "Title for editors" to Homepage obj class.

apepper:
Since [version 1.22.0](https://www.scrivito.com/scrivito-js-1-22-0-released-featuring-language-aware-search-bc64a95b4cb84800) Scrivito provides a way for editors to set a title only for editors. To use it, one has to provide a `contentTitle` attribute of type `string`.

This PR adds `contentTitle` attribute to ~every non-binary obj class~ _Homepage_.

content title in action v3

### DIFF
--- a/src/Objs/Homepage/HomepageEditingConfig.js
+++ b/src/Objs/Homepage/HomepageEditingConfig.js
@@ -20,6 +20,9 @@ Scrivito.provideEditingConfig("Homepage", {
   attributes: {
     ...defaultPageEditingConfigAttributes,
     ...metadataEditingConfigAttributes,
+    contentTitle: {
+      title: "Title for editors",
+    },
     logoDark: {
       title: "Dark logo",
       description: "Used with light backgrounds",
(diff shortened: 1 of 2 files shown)


dcsaszar (changes requested):
From past feedback we know that for typical cases the `title` is already suited for editors.
Only page types that use a "SEO optimized" title (with keywords, product names, whatever) may be distracting or unnecessarily long for an editor. (See release note: https://www.scrivito.com/scrivito-js-1-22-0-released-featuring-language-aware-search-bc64a95b4cb84800#label%20meaningful%20to%20editors)

For other page types, adding "Title for editors" to the properties decreases the user experience. The always-present-but-never-used input is considered noise/creates mental load.

I think the Example App should point out that this editor-specific attribute should be used with care. I suggest to only add it to Homepage, but I'd be fine with adding it to LandingPage also.

apepper:
REVIEW (10499e6)

I still used the `_defaultPageEditingConfig` because otherwise the `contentTitle` could not be placed next to the `title`. I guess it's fine, that `Page` can have a `contentTitle` as well.

apepper:
REVIEW (f773579)

dcsaszar (approved):
LGTM.

Regarding the labels we can do a follow-up discussion.

dcsaszar (approved):
👯‍♂️

PR #395, opened 2021-05-06
Make use of the page defined language as html lang attribute.

apepper:
Falls back to "en" if no language is provided.


dcsaszar (approved):
👍

PR #402, opened 2021-05-31
Fix refactoring oversight in EnvironmentPlugin

dcsaszar:
Introduced in 497c14646d1bc53db41c57d9103a175952b52641

`env` actually looks like `{ WEBPACK_BUNDLE: true, WEBPACK_BUILD: true, production: true }`


apepper (approved):
LGTM (af8f3f0)

Good job 🚀 !

PR #411, opened 2021-06-30
Improved custom widget components

apepper:
IconEditorTab and ColumnsEditorTab should both be be usable in a "read-only" mode.

This is checked using [`Scrivito.canWrite()`](https://www.scrivito.com/js-sdk/canWrite).

IconEditorTab was also converted to React Hooks.

Diff best viewed without whitespace changes.

### Screenshots

Read only icon editor
Read only columns editor

### DIFF
--- a/src/Components/ScrivitoExtensions/ColumnsEditorTab.js
+++ b/src/Components/ScrivitoExtensions/ColumnsEditorTab.js
@@ -18,6 +18,8 @@ class ColumnsEditorTab extends React.Component {
   }
 
   render() {
+    const readOnly = !Scrivito.canWrite();
+
     return (
       <div className="scrivito_detail_content">
         <Alignment widget={this.props.widget} />
@@ -114,6 +116,7 @@ class ColumnsEditorTab extends React.Component {
           <GridLayoutEditor
             currentGrid={this.state.currentGrid}
             adjustGrid={this.adjustGrid}
+            readOnly={readOnly}
           />
         </div>
       </div>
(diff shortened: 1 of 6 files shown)


src/Components/ScrivitoExtensions/IconEditorTab/AllIcons.js
```diff
@@ -2,23 +2,23 @@ import * as React from "react";
 import { take } from "lodash-es";
 import fontAwesomeIcons from "./fontAwesomeIcons";
 
+const categoryMap = {};
+fontAwesomeIcons.forEach((icon) =>
```
dcsaszar, line 6: * [ ] FYI This creates a potentially expensive side-effect of importing. IMO acceptable, because it's only imported  by `scrivito_extensions`.

src/Components/ScrivitoExtensions/IconEditorTab/AllIcons.js
```diff
@@ -12,42 +12,29 @@ fontAwesomeIcons.forEach((icon) =>
   })
 );
 
-class AllIcons extends React.Component {
-  constructor(props) {
-    super(props);
+function AllIcons({ setWidgetIcon, currentIcon, hide }) {
+  const [initialRender, setInitialRender] = React.useState(true);
+  React.useEffect(() => {
+    setTimeout(() => setInitialRender(false), 10);
```
dcsaszar, line 18: * [ ] FYI: reading from the top I thought "Maybe the `Note: the initialRender is a performance tweak` comment should go here".

src/Components/ScrivitoExtensions/IconEditorTab.js
```diff
@@ -19,24 +19,28 @@ function IconEditorTab({ widget }) {
...
-          currentIcon={currentIcon}
-          searchValue={searchValue}
-          setWidgetIcon={setWidgetIcon}
-        />
-        <AllIcons
-          currentIcon={currentIcon}
-          hide={searchValue.length}
-          setWidgetIcon={setWidgetIcon}
-        />
+        {Scrivito.canWrite() ? (
```
dcsaszar, line 22: * [ ] Why ternary instead of `&&`?

src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -238,6 +238,10 @@ const Alignment = Scrivito.connect(({ widget }) => {
       </div>
     </React.Fragment>
   );
+
+  function setAlignment(newAlignment) {
```
dcsaszar, line 255: * [ ] FYI naming it `alignment` instead of `newAlignment` would simplify the code a bit

src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -244,148 +244,150 @@ const Alignment = Scrivito.connect(({ widget }) => {
   }
 });
 
-class GridLayoutEditor extends React.Component {
-  constructor(props) {
-    super(props);
+const GridLayoutEditor = Scrivito.connect(
```
dcsaszar, line 264: * [ ] FYI This component was nicely separated from Scrivito so far. I vote for for keeping it like that, by adding a `readOnly` prop.

src/assets/stylesheets/_grid_layout_editor.scss
```diff
@@ -73,7 +75,7 @@
   width: calc(100% - 2px);
 }
 
-.gle-preview:hover > [class*="grid-col-"] {
+.gle-preview.clickable:hover > [class*="grid-col-"] {
```
dcsaszar, line 78: * [ ] FYI in addition to this, we should also disable `.grid-columns:hover` in read-only mode

src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -185,6 +191,13 @@ const Alignment = Scrivito.connect(({ widget }) => {
       break;
   }
 
+  if (Scrivito.canWrite()) {
+    startAlignmentClasses.push("clickable");
+    centerAlignmentClasses.push("clickable");
+    endAlignmentClasses.push("clickable");
+    stretchAlignmentClasses.push("clickable");
```
dcsaszar, line 198: * [ ] FYI the four initial values are all equal. We could refactor to:
  * Calculate conditional initial value once
  * Assign initial value four times

dcsaszar (changes requested):
Nice stuff. See comments for mainly suggestions

apepper:
REVIEW (d4abfda)

src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -22,7 +22,15 @@ class ColumnsEditorTab extends React.Component {
 
     return (
       <div className="scrivito_detail_content">
-        <Alignment widget={this.props.widget} />
+        <Alignment
+          alignment={this.props.widget.get("alignment")}
+          setAlignment={(alignment) => {
+            if (Scrivito.canWrite()) {
```
dcsaszar, line 28: * [ ] FYI I just realized that `canWrite` can be used in a non-connected context. Otherwise this handler should've been implemented with `Scrivito.load`.

dcsaszar (approved):
Nice!

src/Components/ScrivitoExtensions/ColumnsEditorTab.js
apepper: Do we document somewhere which functions need a loading context and which just works?

dcsaszar: In https://www.scrivito.com/js-sdk/load we state:

> CMS data (like any Obj, Widget or Binary) needs to be loaded from the cloud service before it can be used. When rendering content inside a React component, all CMS data is loaded automatically using Scrivito.connect.

So I think it's safe to assume non-CMS data (like user properties) to go without connect/load, except if explicitly documented like here https://www.scrivito.com/currenteditor-deaac83014976a73.

apepper: The example of https://www.scrivito.com/js-sdk/canWrite does include a loading context (implicitly via `Scrivito.registerComponent`). So at least to me it does not seem obvious, that `canWrite` does not need a loading context.

PR #412, opened 2021-07-05
CookieConsent: Treat denied localStorage as "declined" cookie banner.

apepper:
Google Chrome seems to treat direct access to "localStorage" in an iframe in incognito as a "third party cookie" and blocks it (see [1]).

[1] https://github.com/elastic/kibana/issues/87901#issuecomment-758457425

### DIFF
--- a/src/Components/CookieConsentContext.js
+++ b/src/Components/CookieConsentContext.js
@@ -33,5 +33,14 @@ export function useCookieConsent() {
 }
 
 function getCookieConsent() {
-  return localStorage.getItem("CookieConsent") || "undecided";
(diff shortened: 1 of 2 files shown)


dcsaszar (approved):
Works nicely.
However, the underlying issue means that for an editor, the site setting `Cookie consent link` has no visible effect. It would be nice to be able to test it as an editor.

* [ ] How about we add at least a hint in the editor description?

apepper:
REVIEW (b4c17cf)

dcsaszar (approved):
👍

PR #415, opened 2021-08-06
Columns editor: improve behavior in case of concurrent edits

dcsaszar:


apepper:
Thank you for your work. While reviewing your proposed changes I realized that a second case is not covert by you code: If a concurrent editor adds a new widget the column editor could delete this widget by accident.

I did implement a different approach to solve this problem: see https://github.com/Scrivito/scrivito_example_app_js/pull/424.

PR #416, opened 2021-08-20
webpack-dev-server update to 4.0.0 version

kirkol:
Tiny patch. Updated webpack-dev-server to 4.0.0.

**Note**: two depreciation warnings appear (DEP_WEBPACK_DEV_SERVER_CONSTRUCTOR and DEP_WEBPACK_DEV_SERVER_LISTEN). I wasn't able to find a root case :/ Npm installs packages seamlessly and project runs as it should, so I guess it can be a "general" warning even, if we don't use it in the code

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,8 +69,8 @@
         "typescript": "^4.3.2",
         "use-resize-observer": "^7.0.0",
         "webpack": "^5.38.1",
-        "webpack-cli": "^4.7.0",
-        "webpack-dev-server": "^3.11.2",
+        "webpack-cli": "^4.8.0",
+        "webpack-dev-server": "^4.0.0",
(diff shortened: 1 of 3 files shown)


dcsaszar:
Thanks for the PR. Looks good so far. I'll have a look at the deprecation thingies.

dcsaszar (changes requested):
With `NODE_OPTIONS=--trace-deprecation` we can see the warnings are caused by `webpack-cli`.

* [x] Could you please update this, too?

kirkol:
Poof! Webpack-cli updated as well ;) Thank you for finding that! I set trace-deprecation flag in a wrong place and my Node didn't want to say anything xd

dcsaszar (approved):
Great! Thanks a lot!

PR #420, opened 2021-09-10
Enable optimized widget loading

dcsaszar:
This is available since 1.24.0.
I think that all Example App code is compatible with this setting.


apepper (approved):
LGTM (6ac4e89)

Very nice 🚀 ! I checked every `get("myWidgetListAttributeName")`. All of them are in a connected context.

PR #421, opened 2021-09-14
Replace favicon with new logo

apepper:
+ favicon for Safari


dcsaszar (commented):
I'm not sure about the Apple touch icon. This is something that most developers are not aware of, because it has zero visibility for most devs and editors. I'm afraid most projects starting with the Example App will end up with an Apple touch icon they don't expect.

I'd say starting (and ending up) with none is better than ending up with the wrong one.

apepper:
REVIEW (1a78a14)

It actually works to use the same png for both favicon and apple-touch-icon.

apepper:
REVIEW (78aefad)

Favicon support in Safari just seems to be just broken. I can not reliably add a favicon that works in safari.

So I'm going with your initial suggestion: Just don't add a `apple-touch-icon` and only replace the favicon.

PR #424, opened 2021-10-05
Columns editor: Read only mode in case of concurrent edits

### LINKED ISSUE #414: Column editor leads to an error after concurrent widget removal
* Open the widget editor in tab 1
* Using in-place editing, clear one of the columns of the widget in tab 2
* Make a change to the column widths in tab 1

### Expected

The columns are updated, ideally taking the concurrent changes into account.

### Actual

> An error occurred
> invalid value for key 'content' of widget 641ab6b3: no widget definition for '933e283e' found in '_widget_pool': ["933e283e"]
> 

image

### Discussion

The editor stores the initial column content as state. This state is stale after the concurrent update.

apepper:
This is an alternative approach to #415: Instead of still allowing to edit in the columns editor while concurrent changes are taking place, in my eyes it's better to switch the columns editor into readOnly mode.

This ensures that deleted widgets are handled, but also that newly created widgets are not being deleted.

Fixes #414. Closes #415.

### DIFF
--- a/src/Components/ScrivitoExtensions/ColumnsEditorTab.js
+++ b/src/Components/ScrivitoExtensions/ColumnsEditorTab.js
@@ -4,163 +4,150 @@ import Draggable from "react-draggable";
 import { flatten, isEqual, last, take, takeRight, times } from "lodash-es";
 import ColumnWidget from "../../Widgets/ColumnWidget/ColumnWidgetClass";
 
-class ColumnsEditorTab extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      originalContents: props.widget
-        .get("columns")
-        .map((column) => column.get("content")),
-      currentGrid: gridOfWidget(props.widget),
-    };
+function ColumnsEditorTab({ widget }) {
+  const originalContents = React.useMemo(
+    () => widget.get("columns").map((column) => column.get("content")),
+    [widget]
+  );
+  const [currentGrid, setCurrentGrid] = React.useState(gridOfWidget(widget));
 
-    this.adjustGrid = this.adjustGrid.bind(this);
-  }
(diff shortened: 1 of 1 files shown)


src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -149,6 +153,19 @@ function ColumnsEditorTab({ widget }) {
   }
 }
 
+function calculateContents(widget) {
+  return widget.get("columns").map((column) => column.get("content"));
+}
+
+function calculateContentIds(contents) {
+  return flatten(contents.map((content) => content.map((o) => o.id()))).sort();
```
dcsaszar, line 161: * [ ] Flattening and sorting may lead to missed edits. If our goal is to prevent losing (i.e. overriding) data entered by another editor (here: the position and order of widgets), we need a better content equality model.

src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -5,13 +5,17 @@ import { flatten, isEqual, last, take, takeRight, times } from "lodash-es";
...
-  );
+  const [originalContents, originalContentsId] = React.useMemo(() => {
+    const contents = calculateContents(widget);
+
+    return [contents, calculateContentIds(contents)];
+  }, [widget]);
+
+  const readOnly =
+    !Scrivito.canWrite() ||
+    isConcurrentEditDetected(widget, originalContentsId);
```
dcsaszar, line 16: * [ ] Just switching to read-only, without any indication why, will appear to the editor like the editing is broken.

dcsaszar (changes requested):
See comments.

* [ ] If we are able to detect concurrent edits, could we "reset" the editor at this point?

PR #430, opened 2021-12-07
ButtonWidget: block alignment

apepper:
This PR adds a fourth "alignment": "Block level"

block level

### DIFF
--- a/src/Widgets/ButtonWidget/ButtonWidgetComponent.js
+++ b/src/Widgets/ButtonWidget/ButtonWidgetComponent.js
@@ -16,6 +16,10 @@ Scrivito.provideComponent("ButtonWidget", ({ widget }) => {
   const classNames = ["btn"];
   classNames.push(widget.get("style") || "btn-primary");
 
+  if (widget.get("alignment") === "block") {
+    classNames.push("btn-block");
+  }
+
   return (
     <AlignmentWrapper alignment={widget.get("alignment")}>
       <Scrivito.LinkTag to={target} className={classNames.join(" ")}>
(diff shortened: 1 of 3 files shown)


src/Widgets/ButtonWidget/ButtonWidgetEditingConfig.js
```diff
@@ -12,6 +12,7 @@ Scrivito.provideEditingConfig("ButtonWidget", {
         { value: "left", title: "Left" },
         { value: "center", title: "Center" },
         { value: "right", title: "Right" },
+        { value: "block", title: "Block level" },
```
dcsaszar, line 15: * [ ] FYI In the Bootstrap docs there is a technical term and a human readable term:
> block level buttons—those that span the full width of a parent

I think we should prefer the technical term on the tech side and the human term on the editor side:
```ts
{ value: "block", title: "Full width" },
```

src/Widgets/ButtonWidget/ButtonWidgetComponent.js
```diff
@@ -17,21 +17,19 @@ const ButtonWidgetComponent = Scrivito.connect(({ widget }) => {
...
-    return (
-      <div className={`text-${widget.get("alignment")}`}>
-        <ButtonWidgetComponent widget={widget} />
-      </div>
-    );
-  }
-
-  return <ButtonWidgetComponent widget={widget} />;
-});
+function AlignmentWrapper({ alignment, children }) {
```
dcsaszar, line 33: * [ ] FYI Looking at the next commit it becomes clear that the `alignment` setting is not exclusively evaluated here, which makes the naming slightly off. How about `AlignedText`, `TextAlignment`, `TextWrapper` or so?

dcsaszar (approved):
Nice. Two suggestions.

apepper:
REVIEW (54fe5d9)

dcsaszar (approved):

PR #438, opened 2022-01-24
BlogPost: Allow publishedAt to be more precise (date => datetime)

apepper:
Since Scrivito 1.15.0 `datetime` is supported (see https://www.scrivito.com/scrivito-js-1-15-0-released-featuring-dates-of-first-and-most-recent-publish-4f277ce06c513361). This allows to give an exact order for blog posts published on the same day.


apepper:
Closing for now.

PR #439, opened 2022-01-25
Don't override Bootstrap `btn-lg`/`btn-sm` font size

dcsaszar:
Overriding `.btn` unfortunately also overrides `.btn-sm` and `.btn-lg`.
Now we stick with the default Bootstrap button font size of 1rem,
which should equal 16px.

### DIFF
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -341,12 +341,11 @@ small,
   position: initial !important;
 }
 
-/* button adaoptions
+/* button adaptions
 =================================================================== */
 
 .btn {
   border-radius: 0;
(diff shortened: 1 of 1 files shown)


agessler:
It is an issue how you work with bootstrap itself. You can use (find out) all internal variables or just work your way and care for all results. If you want to use different btn-* classes you have to care for it.

If you want to have an closer look.. go into `_buttons.scss` and see the mess. You have to care and find our which variables are important and involved to change something.

When I do a design I decide what I want and I define what I need. Maybe not the best way.. but I don't want to spend weeks of finding out what the bootstrap guys thought when the over-engeneered their code.

PR #443, opened 2022-01-27
[webpack] Fixed css auto-reload issue.

apepper:
Previously css changes where only visible after a manual refresh of the page - quite annoying. Now css is automatically refreshed on save - similar to js.

It looks like, that this is actually an issue of "html-webpack-plugin" with "contenthash". See https://github.com/jantimon/html-webpack-plugin/issues/1638 . This is more a workaround than an actual fix.


dcsaszar (approved):
👍

PR #447, opened 2022-02-09
Replace Scrivito.registerComponent with direct imports

apepper:
`@loadable/component` is used to not increase the `index.js` file size too much.

Renamed folder `ScrivitoExtensions` to `EditorTabs` and removed "auto-import", since all files are now directly imported.


apepper:
After this we probably have to adjust content on scrivito.com as well: https://www.scrivito.com/search?q=ScrivitoExtensions

package.json
```diff
@@ -84,6 +84,6 @@
     "prerender": "SCRIVITO_PRERENDER=true npm run build && node buildNode/prerender_content.js",
     "start": "webpack serve",
     "eslint": "eslint --max-warnings 0 *.js src/",
-    "es-check": "npm run build && es-check es5 --not=build/assets/scrivito_extensions. './build/**/*.js'"
+    "es-check": "npm run build && es-check es5 --not='build/assets/chunk-*' './build/**/*.js'"
```
dcsaszar, line 87: * [ ] This is dangerous. We already use `@loadable/component` for widgets (`ThumbnailGalleryComponent`), and it's highly recommendable for more extensive use. However, these widgets must still be ES-checked because the code is relevant for *visitors*.

src/Components/EditorTabs/IconEditorTab.js
```diff
@@ -5,7 +5,7 @@ import IconComponent from "../Icon";
 import IconSearch from "./IconEditorTab/IconSearch";
 import IconSearchResults from "./IconEditorTab/IconSearchResults";
 
-function IconEditorTab({ widget }) {
+const IconEditorTab = Scrivito.connect(({ widget }) => {
```
dcsaszar, line 8: * [ ] FYI `export default const` would be a more concise notation. So far we're only using the form `export default function` but I don't see why not `const`?

src/Objs/_metadataEditingConfig.js
```diff
@@ -17,6 +19,10 @@ export const metadataInitialContent = {
   robotsIndex: "yes",
 };
 
+const LoadableSocialCardsTab = loadable(() =>
+  import("../Components/EditorTabs/SocialCardsTab")
+);
```
dcsaszar, line 24: I understand the "make it a const first" notation is copied from the docs. However, the docs use the imported components in JSX, so there must be an uppercase variable.
Here, we could make it more compact by inlining. Ideally, we would completely hide the loadable aspect from the reader, by leaving it as an implementation detail to the imported module. But this would probably open the door for accidentally importing the wrong (not loadable) module.

* [ ] FYI consider inlining
* [ ] FYI maybe consider hiding the loadable aspect from configs

dcsaszar (changes requested):
Nice. See comments.

src/Components/EditorTabs/IconEditorTab.js
apepper: If I try to do `export default const IconEditorTab` I get the following eslint error in my editor:

> Parsing error: Only expressions, functions or classes are allowed as the `default` export.

src/Objs/_metadataEditingConfig.js
apepper: > But this would probably open the door for accidentally importing the wrong (not loadable) module.

That is also what I was thinking. With the inlined variant it now becomes quite obvious, that this is not directly imported.

apepper:
REVIEW ([b6a7f27](https://github.com/Scrivito/scrivito_example_app_js/pull/447/commits/b6a7f27793518321e36a9d2fe0531427f6bba9d5))

src/Components/EditorTabs/IconEditorTab.js
dcsaszar: Sorry, I somehow confused `export const` (valid) with `export default const` (invalid).

dcsaszar (approved):
LGTM 🥳

PR #453, opened 2022-02-25
npm overrides

apepper:
Use "overrides" to help npm version 8.3 and newer.

The "overrides" flag allows to make specific changes to (sub-)dependencies (see [1]). It is supported by npm 8.3 and newer (see [2]).

With this change after an "npm i" "npm audit" will print "found 0 vulnerabilities". Previously it printed "6 moderate severity vulnerabilities".

[1] https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides
[2] https://github.com/npm/cli/releases/tag/v8.3.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
   "repository": {
     "private": true
   },
+  "overrides": {
+    "cli-table3@<0.6.1": "^0.6.1"
+  },
   "dependencies": {
     "@babel/core": "^7.16.7",
     "@babel/eslint-parser": "^7.16.5",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2858,9 +2858,9 @@
       }
     },
     "node_modules/are-we-there-yet": {
-      "version": "1.1.5",
-      "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
-      "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
+      "version": "1.1.7",
+      "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.7.tgz",
+      "integrity": "sha512-nxwy40TuMiUGqMyRHgCSWZ9FM4VAoRP4xUYSTv5ImRog+h9yISPbVH7H8fASCIzYn9wlEv4zvFL7uKDMCFQm3g==",
       "dependencies": {
         "delegates": "^1.0.0",
         "readable-stream": "^2.0.6"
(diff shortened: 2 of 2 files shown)


apepper:
FYI: Travis rejects to run further builds without paying them money. Until this is resolved (e.g. by switching to CircleCI) please wait with the review.

package.json
```diff
@@ -10,6 +10,11 @@
   "repository": {
     "private": true
   },
+  "overrides": {
+    "es-check@^5.2.4": {
+      "cli-table3": "^0.6.1"
+    }
+  },
```
dcsaszar, line 17: If I remove this and `npm i`, `cli-table3` is downgraded but there still are no `npm audit` warnings.

* [ ] Please check if we really need this

apepper: I think this is a bug in `npm`. The `overrides` flag seems to be quite new and tools like `npm audit fix` do not (yet) make use of it. If you remove the `overrides` flag as described and reinstall `es-check` (`npm rm es-check && npm i es-check@^5.2.4`) the `npm audit` warning will reappear.

dcsaszar: How did we create the package-lock? If I undo the change to the lockfile and run `npm i` (npm v.8.1.0) the vulnerabilities don't go away.

apepper: `npm rm es-check && npm i es-check@^5.2.4`

Npm still seams to only in same cases make use of the `overrides` feature. E.g. `npm audit fix` does not make use of `overrides` to fix the issue.

dcsaszar: OK I now have it working as described with npm 8.5.2

I see that we expressed the intention in a very specific way: please patch the unsecure `cli-table3` for a specific package and version.

How about a different approach, given we know that a certain `cli-table3` should not be used at all:

```json
  "overrides": {
    "cli-table3": ">=0.6.1"
  },
```
This would make sure every `cli-table3` dependency uses a safe version, leaving dependencies to good versions intact.

* [ ] As this is a new feature I don't know yet which usage is best practice. WDYT?

apepper: I like the idea. But I think npm is not smart enough (yet?). E.g. if a differente package get's installed, that has a dependency on `cli-table3` version `^1.0.0` (which does not exist) npm would now be able to install version `0.6.1` of that package, because that is in the range of `>=0.6.1`. Also if `cli-table3` version `1.0.0` would get released with a breaking API change, all packages, that rely on the 0.x api could break.

So given that, I would prefer to limit `overrides` as much as possible to avoid messing with other packages as well.

dcsaszar: > npm would now be able to install version 0.6.1 of that package

That would be bad. When I tested it, npm preferred the default over override. Is the expected behavior documented somewhere?

apepper: I found the following documentation:
* Official documentation: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides
* First implementation: https://github.com/npm/arborist/pull/354
* Spec?: https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md

dcsaszar: Nice. As the spec (for me) doesn't answer how and if ranges of dependencies and overrides interact, how about making the intention explicit:
```json
  "overrides": {
    "cli-table3@<0.6.1": "^0.6.1"
  },
```

This would match
> it is arguably better in most cases to apply the override too broadly rather than too narrowly.

and clearly express the idea
> Using this feature should be considered a hack in most cases, something that is done temporarily while waiting for a bug to be fixed

apepper:
REVIEW ([ada8c85](https://github.com/Scrivito/scrivito_example_app_js/pull/453/commits/ada8c85c263098f3892c16eee3e4e66258b8275f))

I forced pushed a new version. The first part is already merged with https://github.com/Scrivito/scrivito_example_app_js/pull/461. The second part now incorporates your feedback as well.

apepper:
FYI: If you run `npm i` with an older version of npm (e.g. `8.2.0`), that does not support `overrides` yet the diff looks like this:

```patch
diff --git a/package-lock.json b/package-lock.json
index fe7dfbacc0..b19cc4aa60 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -13780,7 +13780,7 @@
       "integrity": "sha512-3pWfIwKVdIbB/gWmpLloO6iGAXTRi9mcTinPOwvHfzH3BYjOhLgq2XRG3hKtp+F6vBcBXxMgCobUzBAx1d8T4A==",
       "requires": {
         "bluebird": "^3.4.7",
-        "cli-table3": "^0.6.1",
+        "cli-table3": "^0.5.0",
         "colorette": "^1.0.1",
         "fast-levenshtein": "^2.0.6",
         "lodash": "^4.17.14",
@@ -13877,8 +13877,7 @@
       }
     },
     "cli-table3": {
-      "version": "0.6.1",
-      "resolved": "https://registry.npmjs.org/cli-table3/-/cli-table3-0.6.1.tgz",
+      "version": "https://registry.npmjs.org/cli-table3/-/cli-table3-0.6.1.tgz",
       "integrity": "sha512-w0q/enDHhPLq44ovMGdQeeDLvwxwavsJX7oQGYt/LrBlYsyaxyDnp6z3QzFut/6kLLKnlcUVJLrpB7KBfgG/RA==",
       "requires": {
         "colors": "1.4.0",
```

But it will still install the newer version of `cli-table3`. I'm not sure, if this will be annoying to developers or not.

dcsaszar (approved):
LGTM

> If you run npm i with an older version of npm

This is a general issue with npm. The only difference here is that we can choose between an audit warning and a lockfile change.

I'm fine with both, so feel free to merge or close.

PR #458, opened 2022-03-09
Scrivito 1.29.0 + dynamic editing config

apepper:
[Scrivito `1.29.0`](https://www.scrivito.com/scrivito-js-sdk-1-29-0-released-f7dab02769bdb921) enables to dynamically load editing config later on.

### DIFF
--- a/src/Objs/editingConfigs.js
+++ b/src/Objs/editingConfigs.js
@@ -1,0 +1,6 @@
+function importAll(r) {
+  r.keys().forEach(r);
+}
+
+// import all *EditingConfig.js files under src/Objs/
+importAll(require.context("./", true, /EditingConfig\.js$/));
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "reactstrap": "^8.10.0",
         "sass": "^1.49.0",
         "sass-loader": "^12.4.0",
-        "scrivito": "^1.28.0",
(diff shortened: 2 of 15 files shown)


src/Objs/editingConfigs.js
```diff
@@ -0,0 +1,6 @@
+function importAll(r) {
+  r.keys().forEach(r);
+}
+
+// import all *EditingConfig.js files under src/Objs/
+importAll(require.context("./", true, /EditingConfig\.js$/));
```
dcsaszar, line 6: * [ ] FYI this includes the `_*EditingConfig.js` shared helpers. An alternative could be sth like `/\w+\/\w+EditingConfig\.js$/`.

dcsaszar (approved):
Very nice, 4% (20k) gzipped shaved off!

apepper:
FYI: Forced pushed the same commits based on the current `master` in order to make CircleCI run the build.

src/Objs/editingConfigs.js
apepper: Good point. But I think I would stick to the simpler approach here. The `_metadataEditingConfig.js` file for example would also be included in the bundle, if it would be moved in a subfolder of `src/Objs`. And both `_xEditingConfig.js` files are currently used indirectly in the editing config anyway.

dcsaszar (approved):
👍

PR #460, opened 2022-03-10
Enable CircleCI

apepper:


.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+      - run:
+          name: Install correct npm version
+          command: |
+            echo "Node version"
+            node --version
+            echo ""
+            dpkg --compare-versions `npm -v` ge 7.10 || npm i -g npm@^7.10
+            echo ""
+            echo "NPM version"
+            npm --version
```
dcsaszar, line 27: * [ ] In the [docs](https://circleci.com/developer/orbs/orb/circleci/node) this looks simpler. Is there a reason we can't do it like that?

```yaml
    steps:
      - checkout
      - node/install:
          node-version: '16.13'
      - run: node --version
```

.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+      - node/install-packages
+      - run: npm run << parameters.npm-run-command >>
+
+workflows:
```
dcsaszar, line 31: * [ ] Did you consider `run-npm-command`? From the docs:
```yaml
workflows:
  run-npm-command:
    jobs:
      - node/run:
          npm-run: build
```

dcsaszar (approved):
Looks good. Being a CircleCI noob myself, see comments for discussion.

.circleci/config.yml
apepper: I did. But I could not set `environment` for it. That is why I did implement the job `execute-npm`.

.circleci/config.yml
apepper: I did not want to explicitly state the node version. It should be picket up by `.nvmrc`.

Once we're using node 14 as the baseline (which will be soon - support for node 12 ends in April) we can hopefully remove the "check for npm version 7.10" check.

apepper:
REVIEW ([4bb3b13](https://github.com/Scrivito/scrivito_example_app_js/pull/460/commits/4bb3b13326589eb9fc4eafe585f0c61b682fd7c6))

.circleci/config.yml
gwilczynski: [nodejs 14 comes with npm@6 by default](https://github.com/nodejs/node/discussions/37689#discussioncomment-470971) it means you have to keep "check for npm version 7.10" if you want to use it

PR #462, opened 2022-03-18
Drop support for IE11

apepper:
IE 11 is dead really soon (see https://death-to-ie11.com/). Even the upcoming React version 18 will no longer support this browser (see https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#dropping-support-for-internet-explorer). Time to say goodbye!

This PR changes the baseline from `es5` to `es2018`. AFAICS this is the current `web` default of webpack and should be well supported.

`browserslist` will currently use the following targets:

```json
{
  "android": "96",
  "chrome": "94",
  "edge": "96",
  "firefox": "78",
  "ios": "12.2",
  "opera": "81",
  "safari": "13.1",
  "samsung": "14"
}
```

This PR removes also some CSS and font files, that are no longer needed. Please note, that this is not a complete list of all the things that are no longer needed without IE11. A more thorough look will probably reveal more places, that are now obsolete.

Please note, that I did not configure webpack to compile to native esm (e.g. with https://webpack.js.org/configuration/experiments/#experimentsoutputmodule). This seems to be quite a complex task and needs more testing. Also this feature is marked as an experiment in webpack.

The js package size decreased from 888 KB to 822 KB (minified, not gzipped), so roughly by ~7 %. My rudimentary lighthouse tests did not reveal any improved scoring because of a lighter package size.


dcsaszar (commented):
* [x] What about removing polyfills.js?

dcsaszar (approved):
LGTM 👋 🪐

PR #466, opened 2022-04-04
TestimonialSliderWidget: Replaced empty a tag with button tag

apepper:
Prevents "Links are not crawlable" reports by Lighthouse.


dcsaszar (approved):
👍

PR #467, opened 2022-04-04
[scss] Prettier all scss files

apepper:
> npx prettier --write '**/*.@(scss)'


dcsaszar (approved):
LGTM

* [ ] FYI If we replace `// IE9-11 don't support dppx` by `/* IE9-11 don't support dppx */` Prettier can handle it

apepper:
FYI: I reported the issue to Prettier - https://github.com/prettier/prettier/issues/12600

PR #468, opened 2022-04-06
Improve accessibility

apepper:
Follow up of https://github.com/Scrivito/scrivito_example_app_js/pull/466.

### DIFF
--- a/src/Widgets/DividerWidget/DividerWidgetComponent.js
+++ b/src/Widgets/DividerWidget/DividerWidgetComponent.js
@@ -9,7 +9,7 @@ Scrivito.provideComponent("DividerWidget", ({ widget }) => {
   if (widget.get("showLogo") && root) {
     return (
       <div className="divider-widget">
-        <Scrivito.ImageTag content={root.get("dividerLogo")} />
+        <Scrivito.ImageTag content={root.get("dividerLogo")} alt="Logo" />
       </div>
     );
   }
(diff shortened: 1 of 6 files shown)


src/Widgets/TestimonialSliderWidget/TestimonialSliderWidgetComponent.js
```diff
@@ -66,7 +66,11 @@ function sliderSettings(testimonials) {
     infinite: true,
     speed: 500,
     dotsClass: "testimonial-slider-widget--quote-portrait-wrapper",
-    customPaging: (i) => <button>{testimonialAuthorImages[i]}</button>,
+    customPaging: (i) => (
+      <button aria-label="Show testimonial">
```
dcsaszar, line 70: * [ ] We have a better option here: use the `authorName`. This way, a screen reader can tell the individual buttons apart.

src/Widgets/DividerWidget/DividerWidgetComponent.js
```diff
@@ -9,7 +9,7 @@ Scrivito.provideComponent("DividerWidget", ({ widget }) => {
   if (widget.get("showLogo") && root) {
     return (
       <div className="divider-widget">
-        <Scrivito.ImageTag content={root.get("dividerLogo")} />
+        <Scrivito.ImageTag content={root.get("dividerLogo")} alt="Logo" />
```
dcsaszar, line 12: * [ ] Please see https://accessibleweb.com/question-answer/when-should-i-use-a-null-or-empty-alt-tag/
> Decorative images do not need to be announced by the screen reader, so if the alt attribute is empty (alt="", aka a "null" tag) it will not be announced to the user

dcsaszar (changes requested):
Nice. Two possible improvements, see comments

PR #470, opened 2022-04-13
Bootstrap 5

apepper:
This PR replaces bootstrap 4 with bootstrap 5 (and `bootstrap` with `react-bootstrap`).


dcsaszar (changes requested):
Looks good so far. The only issue I noticed is that headline and text no longer align right.
This is visible, for example, on the landing page "magic features" section.
BS5 introduced `text-start` and `text-end` https://getbootstrap.com/docs/5.0/utilities/text/#text-alignment.

* [ ] We should translate or migrate from left/right -> start/end. Note that we have four custom CSS rules for `text-left`/`text-right`.

apepper:
REVIEW ([1ad493d](https://github.com/Scrivito/scrivito_example_app_js/pull/470/commits/1ad493d27272a0dbec53b8cfef9ea56575620ff6))

`text-left`/`text-right` issues should now be resolved.

Also Alex G. and I tried to reduce the final css file again. Sadly the resulting `index.<hex>.css` file is about 20% bigger then on the current `master` (compare https://app.circleci.com/pipelines/github/Scrivito/scrivito_example_app_js/53/workflows/cf880a71-8d85-48f1-8a03-6f1fb971443e/jobs/128?invite=true#step-110-22 with https://app.circleci.com/pipelines/github/Scrivito/scrivito_example_app_js/47/workflows/d3314b0f-54c3-46a0-9bf9-263a7a91b2c7/jobs/110?invite=true#step-110-22). @dcsaszar Any ideas how to improve the resulting css file size?

src/Widgets/IconWidget/IconWidgetComponent.js
```diff
@@ -7,9 +7,11 @@ function IconWidgetComponent({ widget }) {
   const link = widget.get("link");
   const size = widget.get("size");
 
-  if (["center", "right"].includes(widget.get("alignment"))) {
+  const outerClassName = alignmentClassName(widget.get("alignment"));
```
dcsaszar, line 10: * [ ] FYI Since there is no `innerClassName`, how about `wrapperClassName`? Alternatively, `className` may be good enough, as it's obvious what it's used for.

dcsaszar (commented):
LGTM. The changes reveal that we handle the alignment differently in various places.
* [ ] FYI It would be nice if there was a fits-all helper with the most common behavior, where `left` leads to no `className`.

dcsaszar:
> Any ideas how to improve the resulting css file size?

I wouldn't worry about this too much as long as the Lighthouse score is unaffected.
Optimizing the CSS is probably a story in itself where we could look for and adapt an established best practice.

apepper:
REVIEW ([f6fc1cf](https://github.com/Scrivito/scrivito_example_app_js/pull/470/commits/f6fc1cf23caa16bc262dce84f30c36d27d43d992))

src/Widgets/TextWidget/TextWidgetComponent.js
```diff
@@ -1,31 +1,12 @@
...
-  }
-
-  return "";
-}
+import { alignmentClassName } from "../../utils/alignmentClassName";
+
+Scrivito.provideComponent("TextWidget", ({ widget }) => (
+  <Scrivito.ContentTag
+    tag="div"
+    className={alignmentClassName(widget.get("alignment"))}
```
dcsaszar, line 8: * [ ] FYI In contrast to passing it to `WrapIfClassName` this usage of `alignmentClassName` will lead to a `class=""` attribute. We could make the behavior more consistent (in the sense of "the default doesn't render unnecessary stuff") if `alignmentClassName` returned `undefined`/`null`/false.

dcsaszar (approved):
Nice. One more optional cleanup.

apepper:
REVIEW ([c84365a](https://github.com/Scrivito/scrivito_example_app_js/pull/470/commits/c84365aa116602dabae81d397261eb294fcfab8b))

dcsaszar (approved):
👍

PR #471, opened 2022-04-14
small size fixes

agessler:


apepper:
Manually included in #470. Thank you!

PR #473, opened 2022-04-20
Curly only on multi-line

apepper:
The eslint rule `curly` allows "single line if statements" with the "multi-line" option (see https://eslint.org/docs/rules/curly#multi-line for more details).

It basically allows to write a statement like this:

```js
  if (preconditionNotMet) return null;
```

Previously one had to split this statement up into multiple lines:
```js
  if (preconditionNotMet) {
    return null;
  }
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -47,7 +47,7 @@
         ]
       }
     ],
-    "curly": [1, "all"],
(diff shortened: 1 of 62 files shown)


dcsaszar (approved):
Like it!

PR #474, opened 2022-04-20
Npm audit fix

apepper:
Fixes https://github.com/advisories/GHSA-fwr7-v2mv-hh25

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "private": true
   },
   "overrides": {
-    "cli-table3@<0.6.1": "^0.6.1"
+    "cli-table3@<0.6.1": "^0.6.1",
+    "async@<2.0.0": "^2.6.4"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,8 @@
     "private": true
   },
   "overrides": {
-    "cli-table3@<0.6.1": "^0.6.1"
+    "cli-table3@<0.6.1": "^0.6.1",
+    "async@<2.0.0": "^2.6.4"
```
dcsaszar, line 15: * [ ] Can we update to `"es-check": "^6.2.1"`? This would only need a single override of `"lodash@<4.17.21": "^4.17.21"`. (At least that's what I see locally.)

dcsaszar (commented):
See comment

package.json
apepper: => https://github.com/Scrivito/scrivito_example_app_js/pull/475

PR #484, opened 2022-07-08
Convert default exports to named exports.

apepper:
This PR used the help of the codegenerator https://github.com/apepper/defaultToNamedExport .

Diff best viewed without whitespace changes applied.


dcsaszar (approved):
Incredibe! 💟

PR #489, opened 2022-08-03
👋 prerender.io

apepper:
AFAICS "prerender.io" is now only a "contact us" feature by netlify (see [1]). Since Scrivito supports much better ways now on how to prerender your content we can simplify this code.

[1] https://docs.netlify.com/site-deploys/post-processing/prerendering/#external-services


dcsaszar (approved):
Actually I still see the "prerender with Netlify" option. But using the native Scrivito prerendering is the better option anyways.

image

PR #491, opened 2022-08-08
👋 react-bnb-gallery, welcome again react-slick and react-bootstrap

apepper:
This is a preparation to later support react 17 (and 18).

Sadly "react-bnb-gallery" is no longer maintained (if it ever where, see [1]) and does not support react 17 or 18.

To not run into the same issue with yet-another gallery, let's pick one, that we're already using: `react-slick`. For it to work, it needs to be shown in a dialog, which `react-bootstrap` provides. Since `react-slick` is already part of the package (for testimonial) its not wrapped in a `loadable` component.

@agessler thankfully provided nice stylings.

[1] https://github.com/peterpalau/react-bnb-gallery


src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
```diff
@@ -47,15 +46,71 @@ Scrivito.provideComponent("ThumbnailGalleryWidget", ({ widget }) => {
...
+                d="M23.25 24c-.19 0-.38-.07-.53-.22L12 13.06 1.28 23.78c-.29.29-.77.29-1.06 0s-.29-.77 0-1.06L10.94 12 .22 1.28C-.07.99-.07.51.22.22s.77-.29 1.06 0L12 10.94 22.72.22c.29-.29.77-.29 1.06 0s.29.77 0 1.06L13.06 12l10.72 10.72c.29.29.29.77 0 1.06-.15.15-.34.22-.53.22"
+                fillRule="evenodd"
+              />
+            </svg>
+          </button>
+          <Slider
+            dots
+            dotsClass="slick-dots slick-thumb"
+            infinite
+            speed={500}
```
dcsaszar, line 82: * [ ] FYI I think `500` is the default, so we could drop this. Alternatively, the old implementation behaved like `0`, so that could be an alternative.

src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
```diff
@@ -124,19 +179,36 @@ function allTags(images) {
...
-  return {
-    photo: srcUrl,
-    thumbnail: srcUrl,
-    caption: [
-      galleryImageWidget.get("title"),
-      galleryImageWidget.get("subtitle"),
-    ].join(" - "),
-    alt,
-  };
+function SliderNextArrow(props) {
```
dcsaszar, line 182: * [ ] FYI Other components in this file use parameter destructuring. For consistency we could say `{ onClick }` instead of `props`.

src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
```diff
@@ -124,19 +179,36 @@ function allTags(images) {
...
+      <svg
+        viewBox="0 0 18 18"
+        aria-hidden="true"
+        style={{ height: "2.8em", width: "2.8em", fill: "#fff" }}
+      >
+        <path
+          d="M4.29 1.71A1 1 0 1 1 5.71.3l8 8a1 1 0 0 1 0 1.41l-8 8a1 1 0 1 1-1.42-1.41l7.29-7.29z"
+          fillRule="evenodd"
+        />
+      </svg>
```
dcsaszar, line 194: * [ ] Q: Since we include FontAwesome in the app anyways, is there a particular reason why we don't use the icons here?

dcsaszar (changes requested):
I like it. See comments for some suggestions.
Keyboard nav has some issues, too:

* [ ] Keyboard navigation only works after I click inside the modal
* [x] If keyboard nav is active, there are some glitchy borders around images

Also I think we should improve the appearance on mobile, see side-by-side comparison:

* [x] Prevent text/image overlap
* [ ] Use more horizontal screen estate

image

src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
apepper: No particular reason. Good idea to make use of font awesome.

agessler:
@dcsaszar @apepper  I can improve `Prevent text/image overlap` and 
`Use more horizontal screen estate`. Unfortunately I am under pressure with Neoletter MVP tasks. So I will do my part but have to postpone it for at least 1 week.

src/Widgets/ThumbnailGalleryWidget/ThumbnailGalleryWidgetComponent.js
```diff
@@ -39,7 +39,7 @@ Scrivito.provideComponent("ThumbnailGalleryWidget", ({ widget }) => {
         <div className="row thumbnail-gallery-widget--wrapper">
           {images.map((image, imageIndex) => (
             <Thumbnail
-              key={image.id()}
+              key={`${image.id()}${imageIndex}`}
```
dcsaszar, line 42: * [ ] FYI we construct keys inconsistently. Here we don't have the `-` separator, in line 101 we do.

dcsaszar (commented):
Code changes look good so far.

apepper:
REVIEW ([07b9c37](https://github.com/Scrivito/scrivito_example_app_js/pull/491/commits/07b9c3767712669d9af640935ebeae27d0840326))

Dave and I added some "development style" to fix at least the most pressing issues. I think we can merge this now as is. @agessler can maybe make the gallery look even nicer once he has more time.

dcsaszar (approved):

PR #494, opened 2022-08-23
React 18

dcsaszar:


src/App.js
```diff
@@ -14,12 +14,12 @@ import { CookieConsentProvider } from "./Components/CookieConsentContext";
 
 export const helmetContext = {};
 
-export function App() {
+export function App({ callback }) {
```
apepper, line 17: I know that https://github.com/reactwg/react-18/discussions/5 suggest to name this prop `callback`. But I find it to be quite generic and from reading I don't really know what it suppose to do.

How about something like `onRenderCompletedCallback` or something similar?

apepper (commented):
FEEDBACK ([7bc6b33](https://github.com/Scrivito/scrivito_example_app_js/pull/494/commits/7bc6b33254da764dc1b2556febdb1e0f0fb44c25))

Very nice 🚀! It looks like that react 18 + React.Suspense improves the total-blocking-time in lighthouse scores quite a lot.

See one small naming suggestion. Other than that really good stuff!

PR #501, opened 2022-10-20
replace main content id

Kadirumac:


src/Components/Navigation.js
```diff
@@ -168,7 +168,7 @@ export const Navigation = Scrivito.connect(
             <NavigationSection heightClassName={heightClassName} />
             <ScrollToNextSectionLink heightClassName={heightClassName} />
           </Scrivito.BackgroundImageTag>
-          <div id="mainContent" />
+          <div id="nextSection" />
```
apepper, line 171: If I remember correctly we introduced this `div` back in the days, when we used a dedicated React library to to the scrolling. Since https://github.com/Scrivito/scrivito_example_app_js/commit/d9942c965bf8d0451e65e81b90aa7f4ce0b1a4ca `div` here is strictly speaking no longer needed.

How about the following:
* Wrap `<Scrivito.CurrentPage />` into `<div id="mainContent">` as you did in this PR.
* Remove this line.
* Leave `ScrollToNextSectionLink` to link to `#mainContent`.

This way we only have one div to jump to and the ID stays the same.

The only downside I see, is that `ScrollToNextSectionLink` now relies on `<App>` to render a `mainContent` id. But it previously relied on `<Navigation>` to do just that.

apepper (commented):
FEEDBACK ([4fc35e0](https://github.com/Scrivito/scrivito_example_app_js/pull/501/commits/4fc35e04644f7739fb16fdb5f762493a613cf510))

Really nice 🚀. Thank you for your contribution and the level of detail you look at.

I have one suggestion to streamline the code even further (see comment below).

apepper (approved):
LGTM ([33435c0](https://github.com/Scrivito/scrivito_example_app_js/pull/501/commits/33435c062ecb050e5c9085d00f654eaa31fa9c2e))

Nice 🚀.

Thank you for your contribution!

PR #507, opened 2022-12-12
add binary check for search results

Kadirumac:


src/Objs/SearchResults/SearchResultItem.js
```diff
@@ -12,6 +12,7 @@ const PreviewImage = Scrivito.connect(({ item }) => {
     item.get("navigationBackgroundImage") ||
     item.get("titleImage") ||
     item.get("image");
+  if (image instanceof Scrivito.Binary) return null;  
   if (!image || isVideoObj(image)) return null;
```
apepper, line 16: Please check, that image is (not) an Obj. Otherwise every other attribute time other than `binary` would also fail this component. And to make the code flow clearer I would split each guard to its own line:

```suggestion
  if (!image) return null;
  if (!(image instanceof Scrivito.Obj)) return null;
  if (isVideoObj(image)) return null;
```

apepper (changes requested):
FEEDBACK ([034df32](https://github.com/Scrivito/scrivito_example_app_js/pull/507/commits/034df325fe4a48d192c5a7f0450e54fda2d5095b))

Very nice 🚀! Please see my comment for further improvements.

Kadirumac:
@apepper  Your suggestion is added. ✅

apepper (approved):
LGTM ([71cc449](https://github.com/Scrivito/scrivito_example_app_js/pull/507/commits/71cc449168f7d6cef0e7f88ed652626081c83a4a))

Very nice 🚀. Thank you for your contribution 🎁!

src/Objs/SearchResults/SearchResultItem.js
```diff
@@ -12,8 +12,9 @@ const PreviewImage = Scrivito.connect(({ item }) => {
     item.get("navigationBackgroundImage") ||
     item.get("titleImage") ||
     item.get("image");
-  if (image instanceof Scrivito.Binary) return null;  
-  if (!image || isVideoObj(image)) return null;
+  if (!image) return null;
+  if (!(image instanceof Scrivito.Obj)) return null;
```
phorsuedzie, line 16: FYI: To handle the "falsy" case separately is not actually required. Worst case (works also for `null`, `false`, `""`, and `0` (and likely for any other falsy JS thingy which is never passed here actually)):
```
undefined instanceof Object 
// => false
```

PR #508, opened 2022-12-19
Allow sorting children for all types of objects by default

gwilczynski:
The order of the child pages is defined by the `childOrder` attribute of the parent page. This attribute is assumed to be of the [referencelist](https://www.scrivito.com/custom-attribute-types-a0a749848457286c) type. If it’s missing in the corresponding [page class definition](https://www.scrivito.com/js-sdk/createObjClass), then the children are ordered randomly, and it will not be possible for editors to adjust this order.

I think it would make sense to allow sorting children for all types of objects by default. Especially if UI is ready for this and Editor can confused why she/he can't use this feature.

image

### DIFF
--- a/src/Objs/Homepage/HomepageObjClass.js
+++ b/src/Objs/Homepage/HomepageObjClass.js
@@ -8,7 +8,6 @@ export const Homepage = Scrivito.provideObjClass("Homepage", {
     showAsLandingPage: "boolean",
     cookieConsentLink: "link",
     contentTitle: "string",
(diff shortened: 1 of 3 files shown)


apepper (approved):
LGTM ([a856a14](https://github.com/Scrivito/scrivito_example_app_js/pull/508/commits/a856a147aab3bf7a1a1a12391e36df82f0997df7))

Very nice 🚀! Thank you for your contribution.

PR #509, opened 2023-01-03
More robust attributes

apepper:
This is a continuation of #507.

As we have seen, there are projects out that, that used e.g. a `image` attribute of type `widgetlist` or `binary`. In those cases, the example app should not crash, but function correctly.


src/utils/getMetadata.js
```diff
@@ -80,7 +80,7 @@ function firstImageUrlForAttributes(obj, attributes) {
     if (url) return;
 
     const binary = obj.get(attribute);
-    if (isVideoObj(binary)) return;
+    if (!isImageObj(binary)) return;
```
dcsaszar, line 83: * [ ] FYI we could optimistically name the const `image` like in SearchResultItem

dcsaszar (approved):
LGTM

apepper:
REVIEW ([f78b578](https://github.com/Scrivito/scrivito_example_app_js/pull/509/commits/f78b5784d5998c24770855c636e67ef37c2240b6))

PR #512, opened 2023-01-19
fix component order

markus-codewrangler:
wrong component order leads to error in older versions of [ErrorBoundary.js](https://github.com/Scrivito/scrivito_example_app_js/blob/907d43020c9a91f531cb8f242e5aa20ac406a673/src/Components/ErrorBoundary.js)

see discussion in https://infopark.slack.com/archives/C5LGN1D42/p1674119386473099

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -16,9 +16,9 @@ export const helmetContext = {};
 
 export function App({ appWrapperRef }) {
   return (
-    <ErrorBoundary>
-      <CookieConsentProvider>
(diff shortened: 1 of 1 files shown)


apepper (approved):
LGTM ([292aa97](https://github.com/Scrivito/scrivito_example_app_js/pull/512/commits/292aa977a74821a961b26c04a2941a9f4af3ec8f))

Nice 🚀. Thank you for your contribution!

PR #516, opened 2023-03-03
HeadlineWidget: Improved automatic anchors.

apepper:
kebabCase will try to covert camelCase to camel-case:

```js
_.kebabCase('fooBar');
// => 'foo-bar'
```

Source: https://lodash.com/docs/4.17.15#kebabCase

This can have undesirable effects, if one inputs "normal" text. E.g. 

```js
_.kebabCase("Mapping sites and URLs dynamically")
// => "mapping-sites-and-ur-ls-dynamically"
```

With this change the anchor now looks like this:
```js
// => "mapping-sites-and-urls-dynamically"
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -66,6 +66,7 @@
         "scrivito": "^1.31.0",
         "scroll-to-fragment": "^1.0.12",
         "slick-carousel": "^1.6.0",
+        "speakingurl": "^14.0.1",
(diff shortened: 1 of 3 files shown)


dcsaszar (approved):
Looks good for a workaround.
We should track https://github.com/lodash/lodash/issues/4726, maybe at some point it will be fixed upstream.

PR #518, opened 2023-03-10
Scrivito 1.32.0

apepper:
See https://www.scrivito.com/scrivito-js-sdk-1-32-0-released-6dc70e6597de3661 for all the greatness of the release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "react-slick": "^0.29.0",
         "sass": "^1.55.0",
         "sass-loader": "^13.0.2",
-        "scrivito": "^1.31.0",
+        "scrivito": "^1.32.0",
         "scroll-to-fragment": "^1.0.12",
         "slick-carousel": "^1.6.0",
         "speakingurl": "^14.0.1",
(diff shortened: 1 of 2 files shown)


src/Widgets/FactWidget/FactWidgetEditingConfig.js
```diff
@@ -4,6 +4,20 @@ import factWidgetIcon from "../../assets/images/fact_widget.svg";
...
+        toolbar: ["subscript", "superscript"],
+      },
+    },
+    value: {
+      options: {
+        allowedTags: ["sub", "sup"],
+        toolbar: ["subscript", "superscript"],
+      },
+    },
+  },
```
dcsaszar, line 20: * [ ] Basically same as above. Customers typically ask for sub/sup in headlines. I think `string` is the better default everywhere else. If needed, devs can easily change from string to HTML on their own.

src/Widgets/HeadlineWidget/HeadlineWidgetClass.js
```diff
@@ -2,7 +2,7 @@ import * as Scrivito from "scrivito";
 
 export const HeadlineWidget = Scrivito.provideWidgetClass("HeadlineWidget", {
   attributes: {
-    headline: "string",
+    headline: "html",
```
dcsaszar, line 5: * [ ] If we do this, we must update the `id={speakingUrl(widget.get("headline"))}` part.

dcsaszar (changes requested):
Looks good aside from some rough edges.

See feedback for using the new `allowedTags` feature. I'm not even fully convinced this should be the new default for headlines since it adds a bit of complexity to both the app code and the editing UI. So far, this feature was only requested by large-scale customer projects. Maybe we should just leave the default as is?

apepper:
I forced pushed [798d310](https://github.com/Scrivito/scrivito_example_app_js/pull/518/commits/798d3108d775e9ba0e068a2434464a585c06cd0c) . With that only the upgrade is included

PR #521, opened 2023-05-23
Use correct comparison value for uiContext()

markus-codewrangler:
see https://www.scrivito.com/uicontext-86491ad46e7205ad

as requested by @apepper

### DIFF
--- a/src/Components/ScrivitoExtensions/ColumnsEditorTab.js
+++ b/src/Components/ScrivitoExtensions/ColumnsEditorTab.js
@@ -7,7 +7,7 @@ import { ColumnWidget } from "../../Widgets/ColumnWidget/ColumnWidgetClass";
 export function ColumnsEditorTab({ widget }) {
   const includedWidgetIds = calculateContentIds(calculateContents(widget));
   const uiContext = Scrivito.uiContext();
-  if (uiContext === null) return null;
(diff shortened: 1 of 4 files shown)


src/Components/ScrivitoExtensions/ColumnsEditorTab.js
```diff
@@ -7,7 +7,7 @@ import { ColumnWidget } from "../../Widgets/ColumnWidget/ColumnWidgetClass";
 export function ColumnsEditorTab({ widget }) {
   const includedWidgetIds = calculateContentIds(calculateContents(widget));
   const uiContext = Scrivito.uiContext();
-  if (uiContext === undefined) return null;
+  if (uiContext === null) return null;
```
apepper, line 10: Seeing this and the documentation of `uiContext` I think we can simplify this even further:

```suggestion
  if (uiContext) return null;
```

Even if `uiContext` would be an empty string the function would now not render instead of a rendering a dysfunctional `scrivito_` css classname.

apepper (approved):
LGTM ([29c85f7](https://github.com/Scrivito/scrivito_example_app_js/pull/521/commits/29c85f75b657d28508b18c6633519d724720099c))

Really nice. I found just one more optimization (see comment).

PR #32, opened 2018-02-12
Add Google analytics support

code-constructor:
It's now possible to enable Google Analytics. Just input your tracking ID in the "Site settings":

google analytics

PR #523, opened 2023-06-01
Removed "Universal Analytics" Google Analytics

apepper:
"Universal Analytics" will soon be dysfunctional (see [1]). Cookie consent and `src/tracking.js` are still there, so that people can easily add their tracking code.

[1] https://support.google.com/analytics/answer/11583528

### DIFF
--- a/src/tracking.js
+++ b/src/tracking.js
@@ -1,9 +1,1 @@
-import "autotrack/lib/plugins/url-change-tracker";
-
-const googleScript = document.createElement("script");
-googleScript.setAttribute("async", true);
-googleScript.setAttribute(
-  "src",
-  "https://www.google-analytics.com/analytics.js"
-);
-document.head.appendChild(googleScript);
+// Paste your tracking code (e.g. Google Analytics) here.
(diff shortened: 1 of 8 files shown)


src/tracking.js
```diff
@@ -1,9 +1 @@
-import "autotrack/lib/plugins/url-change-tracker";
-
-const googleScript = document.createElement("script");
-googleScript.setAttribute("async", true);
-googleScript.setAttribute(
-  "src",
-  "https://www.google-analytics.com/analytics.js"
-);
-document.head.appendChild(googleScript);
+// Paste your tracking code (e.g. Google Analytics) here.
```
dcsaszar, line 1: * [ ] As we introduced this script solely for GA (#32) I think it's cleaner to drop it now.

dcsaszar (changes requested):
Looks good so far.

src/tracking.js
apepper: If we drop it there is also no need for the cookie consent. But I would like to keep the cookie consent code around, so that it is fairly easy to use it.

dcsaszar: How about applying the consent at some existing point, like Gmaps or videos, where we currently always use the no-cookies version?

FYI The red flag for me here was the comment plus the fact that the GA code did not (only) live here, which may be misleading.

dcsaszar (approved):

PR #524, opened 2023-06-05
Fix property text inputs on Firefox

dcsaszar:
| before | after |
| -- | -- |
| image |  image |

This lead to bad UX if the input contains words longer than the available width, which happens most frequently in sidebar editing. Chrome automatically uses this style for contenteditable. Now, all browsers behave the same.


apepper (approved):
LGTM ([f0b20ce](https://github.com/Scrivito/scrivito_example_app_js/pull/524/commits/f0b20ce512c5281542b252ad19443c49f4a24ded))

Nice 🚀!

PR #530, opened 2023-06-19
Bugfix: Render hidden fields as a dedicated component

### LINKED ISSUE #528: Hidden Field Widget in Form shows error
In a Form Widget AND NOT in a Cloumn Widget I am able to add Hidden Fields. There is no view for it so one gets "Widget could not be rendered due to application error.".
Though editors are supposed to add the Hidden Field in the FormWidget Properties, they can add them in the Form container. Instead of the error above they should somehow get a hint to use the properties.

apepper:
Fixes #528

### DIFF
--- a/src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
+++ b/src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
@@ -3,7 +3,7 @@ import formHiddenFieldWidgetIcon from "../../assets/images/form_hidden_field_wid
 import { customFieldNameValidation } from "../FormContainerWidget/utils/validations/customFieldNameValidation";
 
 Scrivito.provideEditingConfig("FormHiddenFieldWidget", {
-  title: "Hidden Field",
+  title: "Hidden Form Field",
   thumbnail: formHiddenFieldWidgetIcon,
   attributes: {
     customFieldName: { title: "Field name" },
@@ -37,7 +37,10 @@ Scrivito.provideEditingConfig("FormHiddenFieldWidget", {
     },
   ],
   titleForContent: (widget) =>
-    `Hidden: ${[widget.get("customFieldName"), widget.get("hiddenValue")]
+    `Hidden Form Field: ${[
+      widget.get("customFieldName"),
+      widget.get("hiddenValue"),
+    ]
       .filter((e) => e)
       .join(" - ")}`,
 });
--- a/src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
+++ b/src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
@@ -30,8 +30,7 @@ Scrivito.provideEditingConfig("FormHiddenFieldWidget", {
       }
 
       return {
-        message:
-          "Usually hidden fields should be added in the widget properties of the form container.",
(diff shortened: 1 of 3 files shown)


dcsaszar (approved):
Looks good but doesn't fix #528 

> they should somehow get a hint to use the properties.

* [ ] How about:
```jsx
(widget) =>
      widget.container().get("hiddenFields").includes(widget)
        ? undefined
        : "Please use the form properties for hidden fields",
```

apepper:
After talking verbally we agreed on the following:
* We'll add a validation with a "info", that this is not the best place to put a hidden field.
* We'll update the "titleForContent" of the HiddenFieldWidget to also include, that this is a hidden field widget.

apepper:
REVIEW ([5e17a14](https://github.com/Scrivito/scrivito_example_app_js/pull/530/commits/5e17a144969105fcb9037d5d4753df0dcac04702))

src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
```diff
@@ -19,7 +19,10 @@ Scrivito.provideEditingConfig("FormHiddenFieldWidget", {
   },
   validations: [customFieldNameValidation],
   titleForContent: (widget) =>
-    [widget.get("customFieldName"), truncate(widget.get("hiddenValue"))]
-      .filter((e) => e)
-      .join(": "),
+    truncate(
+      `Hidden: ${[widget.get("customFieldName"), widget.get("hiddenValue")]
```
dcsaszar, line 42: * [ ] The title is not in line with other form fields. E.g. "Form Checkbox", "Form Input Field", "Form Button". Thus "Hidden Form Field" would be more consistent.

src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
```diff
@@ -19,7 +19,10 @@ Scrivito.provideEditingConfig("FormHiddenFieldWidget", {
...
+        .filter((e) => e)
+        .join(" - ")}`,
+      { length: 36 }
```
dcsaszar, line 45: * [ ] The UI already trims titles where needed. In places where there's enough room, the full title is better.

src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetEditingConfig.js
```diff
@@ -17,7 +17,26 @@ Scrivito.provideEditingConfig("FormHiddenFieldWidget", {
...
+          .get("hiddenFields")
+          .map((w) => w.id())
+          .includes(widget.id())
+      ) {
+        return;
+      }
+
+      return {
+        message:
+          "Usually hidden fields should be added in the widget properties of the form container.",
```
dcsaszar, line 35: * [ ] "Usually" implies that there are exceptions. I think this is more confusing that helpful. Also, "widget" may be left out since it's always a widget. In the UI we call it "Form", not "Form Container". How about "Hidden fields should be added in the properties of the form.". Also not sure about in/to/at here.

dcsaszar (changes requested):
Looks good so far. See comments for feedback regarding wording and formatting.

apepper:
REVIEW ([f448405](https://github.com/Scrivito/scrivito_example_app_js/pull/530/commits/f448405e7f62ac2338d59dac01e95c90f39aa89a))

PR #535, opened 2023-07-24
Neoletter || Form builder || Add subscription checkbox type 

dawgoc:
Neoletter has introduced partial (beta) support for "subscription". To make this feature compatible with the example app "Form Builder," I've implemented the first change by introducing a new checkbox type "subscription". 

As the "subscription" feature is still in its development stage and not yet documented, it remains concealed for standard users. Following the approach taken during the initial stages of Form Builder, I have opted to hide the "subscription" functionality under a special environment variable. 

`ENABLE_NEOLETTER_FORM_BUILDER_SUBSCRIPTION_FEATURE`

This approach ensures that the feature is accessible for testing and refinement while remaining hidden from regular users until it is fully prepared for deployment.

This initial pull request sets the groundwork for the integration of Neoletter subscription feature with Form Builder.

Screenshot 2023-07-24 at 14 44 39
Screenshot 2023-07-24 at 14 44 43

cc @jfienhold @7ack7

### DIFF
--- a/src/Widgets/FormCheckboxWidget/FormCheckboxWidgetClass.js
+++ b/src/Widgets/FormCheckboxWidget/FormCheckboxWidgetClass.js
@@ -4,7 +4,16 @@ export const FormCheckboxWidget = Scrivito.provideWidgetClass(
   "FormCheckboxWidget",
   {
     attributes: {
-      type: ["enum", { values: ["custom", "accept_terms"] }],
+      type: [
+        "enum",
+        {
+          values: ["custom", "accept_terms"].concat(
+            process.env.ENABLE_NEOLETTER_FORM_BUILDER_SUBSCRIPTION_FEATURE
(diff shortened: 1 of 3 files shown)


apepper (commented):
FEEDBACK ([850ff3d](https://github.com/Scrivito/scrivito_example_app_js/pull/535/commits/850ff3de05286f078fae711b9ba023dd3d595dd9))

Very nice 🚀. Thank you for your contribution!

The only thing that I would change is the order of the options. I think an order of "Accept terms", "Subscription" and "Custom" would be nicer.

apepper (approved):
LGTM ([be53e50](https://github.com/Scrivito/scrivito_example_app_js/pull/535/commits/be53e505371f4fd9c53fa2c4b1cb686e1239b27d))

Nice 🚀!

PR #542, opened 2023-08-21
Neoletter | Form Builder | Subscription as a hidden field 

dawgoc:
Related to https://github.com/Scrivito/scrivito_example_app_js/pull/535

To make the subscription process more straightforward, we've made a change in the Form Builder. Previously, users had to actively check a specific checkbox to subscribe. We realized this step might not be necessary for all types of subscription forms. To simplify the process, we've incorporated the subscription action as a hidden field within the form. This means that when users fill out the form, the subscription process is initiated without the need for the user to check an additional checkbox.

Currently, subscription feature-related fields are activated when the ENABLE_NEOLETTER_FORM_BUILDER_SUBSCRIPTION_FEATURE environment variable flag is set.

Without flag: 
Screenshot 2023-08-21 at 14 27 35

With flag: 
Screenshot 2023-08-21 at 14 13 54
Screenshot 2023-08-21 at 14 13 59
Screenshot 2023-08-21 at 14 14 10

### DIFF
--- a/src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetClass.js
+++ b/src/Widgets/FormHiddenFieldWidget/FormHiddenFieldWidgetClass.js
@@ -5,6 +5,16 @@ export const FormHiddenFieldWidget = Scrivito.provideWidgetClass(
   {
     onlyInside: "FormContainerWidget",
     attributes: {
+      type: [
+        "enum",
+        {
+          values: ["custom"].concat(
+            process.env.ENABLE_NEOLETTER_FORM_BUILDER_SUBSCRIPTION_FEATURE
+              ? ["subscription"]
+              : []
+          ),
+        },
+      ],
       customFieldName: "string",
       hiddenValue: "string",
     },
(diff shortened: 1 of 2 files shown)


apepper:
FYI: I will not be able to review this PR in the near future. My queue is quite full with high priority tasks.

jfienhold:
> FYI: I will not be able to review this PR in the near future. My queue is quite full with high priority tasks.

Would be cool, if we could get this merged 😎

apepper:
As far as I know Amier is working on an NPM package to incorporate the form builder. To me this would be the right place to add such functionality.

aea-JR:
I adapted it to the Scrivito-Neoletter-Form-Widgets package and also reviewed it. The code appears to be in good shape, so perhaps you could approve it easily? Or perhaps someone else from your team could review it as well? I'm willing to take responsibility if any issues arise.

The situation is that there are existing customers who want that feature but aren't currently using the package. They've made some modifications to their form widgets that might not function properly if they switch to using the package.

apepper:
@aea-JR : Thank you for your offer. But I think it would be useful if:

1. The example app uses `scrivito-neoletter-form-widgets` directly.
2. `scrivito-neoletter-form-widgets` provides a way to customers/projects to provide custom code.

Could you create a PR that removes the existing form widgets from the example app and use `scrivito-neoletter-form-widgets` directly?

apepper:
Closing in favor of incorporating `scrivito-neoletter-form-widgets` instead.

PR #551, opened 2023-11-27
Scrivito 1.38.0

apepper:


src/config/scrivito.js
```diff
@@ -3,6 +3,7 @@ import * as Scrivito from "scrivito";
 export function configureScrivito(options) {
   const config = {
     adoptUi: true,
+    autoConvertAttributes: true, // Details: https://www.scrivito.com/automatic-type-conversions-0cff08a7795d5d3c
```
dcsaszar, line 6: * [ ] We should leave out the comment. Every option can be looked up in the docs, including links to details if needed.

dcsaszar (commented):
See comment

PR #554, opened 2024-02-07
Simplify htmlFor

dcsaszar:
The removed code was a workaround for a Scrivito in-place editing issue fixed in 1.34.0.

### DIFF
--- a/src/Widgets/FormCheckboxWidget/FormCheckboxWidgetComponent.js
+++ b/src/Widgets/FormCheckboxWidget/FormCheckboxWidgetComponent.js
@@ -6,11 +6,6 @@ import { getFieldName } from "../FormContainerWidget/utils/getFieldName";
 Scrivito.provideComponent("FormCheckboxWidget", ({ widget }) => {
   const id = `form_checkbox_widget_${widget.id()}`;
 
-  const labelOptions = {};
-  if (!Scrivito.isInPlaceEditingActive()) {
-    labelOptions.htmlFor = id;
-  }
-
   return (
     <div className="form-check mb-3">
       <input
@@ -26,7 +21,7 @@ Scrivito.provideComponent("FormCheckboxWidget", ({ widget }) => {
         content={widget}
         attribute="label"
         tag="label"
-        {...labelOptions}
+        htmlFor={id}
       />
       {widget.get("required") ? (
         <OverlayTrigger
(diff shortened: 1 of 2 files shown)


src/Widgets/FormCheckboxWidget/FormCheckboxWidgetComponent.js
```diff
@@ -26,7 +21,7 @@ Scrivito.provideComponent("FormCheckboxWidget", ({ widget }) => {
         content={widget}
         attribute="label"
         tag="label"
-        {...labelOptions}
+        htmlFor={id}
```
apepper, line 24: With that the checkbox comes selected at least for a fraction of a second:


https://github.com/Scrivito/scrivito_example_app_js/assets/86275/de1e0db8-fe12-4b5e-ad4a-ed07aea310cb

I preferred the previous behavior, where the checkbox does not change on click at all.

apepper (commented):
FEEDBACK (841185095f5dba06da21c9141e2435c31ef59e6b)

`FormInputFieldWidgetComponent.js` looks good, but `FormCheckboxWidgetComponent.js` is not ideal (see comments).

src/Widgets/FormCheckboxWidget/FormCheckboxWidgetComponent.js
dcsaszar: > FormInputFieldWidgetComponent.js looks good, but FormCheckboxWidgetComponent.js is not ideal

Both behave the same. The difference is only in app styles. Text inputs in this app currently don't have dedicated styles while `:active` (mouse button pressed on the input or label), while checkbox inputs do.

> I preferred the previous behavior, where the checkbox does not change on click at all.

What is the reason to hide the `:active` style in in-place-editing?
One could argue that applying this style in in-place-editing is more WYSIWYG.
For example, an editor would see the difference in styling of text + input vs. label + input without having to switch to preview mode.

apepper: I see. Then let's keep it like you proposed.

apepper (approved):
LGTM ([8411850](https://github.com/Scrivito/scrivito_example_app_js/pull/554/commits/841185095f5dba06da21c9141e2435c31ef59e6b))

Nice 🚀!

PR #572, opened 2024-07-23
Fix ESLint if SCRIVITO_TENANT is unset

dcsaszar:
Without this, `npm run eslint` fails with
```
✖ 3499 problems (3499 errors, 0 warnings)
```

The README doesn't mention `.env` with `SCRIVITO_TENANT` as a prerequisite for developing, so
I think `ESLint` should work out of the box.


apepper (approved):
LGTM (b166f35c5aed752771cf72871137bb9f3a618ba8)

Nice 🚀!

PR #592, opened 2025-09-03
[readme] Add deprecation warning

apepper:
This repository is deprecated in favor of the new [Scrivito Portal App](https://github.com/Scrivito/scrivito-portal-app).

After merge, we should follow the instructions of https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories.

### DIFF
--- a/readme.mdown
+++ b/readme.mdown
@@ -1,6 +1,6 @@
 # Scrivito Example App JS
 
-> **⚠️ DEPRECATED**: This repository is deprecated and is no longer maintained. Please use the new [Scrivito Portal App](https://github.com/Scrivito/scrivito-portal-app) instead.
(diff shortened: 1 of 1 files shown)


readme.mdown
```diff
@@ -1,5 +1,7 @@
 # Scrivito Example App JS
 
+> **⚠️ DEPRECATED**: This repository is deprecated and is no longer maintained. Please use the new [Scrivito Portal App](https://github.com/Scrivito/scrivito-portal-app) instead.
+
```
dcsaszar, line 4: * [ ] How about a more encouraging description?
  * "deprecated" can be misunderstood. This repo is still a good historic reference for how things can be done and nothing in it is deprecated, as Scrivito is 100% backwards compatible.
  * Maybe start with the fact that development has moved?
  * We could also list some of the new features of the Portal App so people get curious to have a look. The notice will also stick out more, and push the old README stuff down.
  * Maybe drop the warning icon? I think GitHub already highlights archived repos.
  * Check with Andreas for wording?

dcsaszar (commented):
See comment

dcsaszar (approved):

dcsaszar (approved):
👍

